PR #12, opened 2019-06-10
- code review changes

LeonardoGonzales:


uniprotkb-rest/src/test/java/uk/ac/ebi/uniprot/api/uniprotkb/controller/UniprotKBSearchControllerIT.java
```diff
@@ -759,10 +352,417 @@ public void searchForDeletedInactiveEntriesReturnItself() throws Exception {
...
+            case mass:
+                value = "[* TO *]";
+                break;
+            case proteome:
+                value = "UP000000000";
+                break;
+            case annotation_score:
+                value = "5";
+                break;
+        }
```
eddturner: should we have a default case, which throws some kind of runtimeexception to show that something is wrong?

uniprotkb-rest/src/test/java/uk/ac/ebi/uniprot/api/uniprotkb/controller/UniprotKBSearchControllerIT.java
```diff
@@ -759,10 +352,417 @@ public void searchForDeletedInactiveEntriesReturnItself() throws Exception {
...
+
+    @Override
+    protected String getFieldValueForValidatedField(SearchField searchField) {
+        UniProtField.Search search = UniProtField.Search.valueOf(searchField.getName());
+        log.debug("FIELD_NAME: ", search.getName());
+        String value = "";
+        if (searchField.getName().startsWith("ftlen_")) {
+            value = "[* TO *]";
+        }
+        switch (search) {
```
eddturner: would it make sense to have constants for some of the possible values that are used in the tests? E.g., private static final String ACCESSION = "P21802"; which is used on line 301, 602, 641 and maybe other places? Same point could go for other values seen in this switch statement.

The point of the suggestion is really only to save time in future if someone wants to refactor something.

uniprotkb-rest/src/test/java/uk/ac/ebi/uniprot/api/uniprotkb/controller/UniprotKBSearchControllerIT.java
LeonardoGonzales: I can not use the default here... because if the result of this method is empty I do extract things in the abstract layer... This is that calls this method:

    private String getFieldValueForField(SearchField searchField) {
        String value = getFieldValueForValidatedField(searchField);
        if (value.isEmpty()) {
            if (searchField.hasValidValue("*")) {
                value = "*";
            } else if (searchField.hasValidValue("true")) {
                value = "true";
            }
        }
        return value;
    }

uniprotkb-rest/src/test/java/uk/ac/ebi/uniprot/api/uniprotkb/controller/UniprotKBSearchControllerIT.java
LeonardoGonzales: done... tks

PR #19, opened 2019-07-17
Update the cross ref service to use BasicSearchController and BaseSer…

sahmad-ebi:
…vice Classes


support-data-rest/src/test/java/uk/ac/ebi/uniprot/crossref/controller/CrossRefGetIdControllerIT.java
```diff
@@ -0,0 +1,179 @@
...
+
+    static class CrossRefGetIdContentTypeParamResolver extends AbstractGetIdContentTypeParamResolver {
+
+        @Override
+        public GetIdContentTypeParam idSuccessContentTypesParam() {
+            return GetIdContentTypeParam.builder()
+                    .id(ACCESSION)
+                    .contentTypeParam(ContentTypeParam.builder()
+                            .contentType(MediaType.APPLICATION_JSON)
+                            .resultMatcher(jsonPath("$.link_type", is("Explicit")))
```
eddturner: Would it be possible for these names to be camel case (same for the ones below)?

sahmad-ebi: Will fix as a separate issue fir disease and cross ref responses.

PR #20, opened 2019-07-19
Write a rest endpoint to download diseases

sahmad-ebi:


support-data-rest/src/test/java/uk/ac/ebi/uniprot/disease/DiseaseSearchControllerIT.java
```diff
@@ -309,6 +309,10 @@ protected SearchContentTypeParam searchSuccessContentTypesParam() {
                             .contentType(UniProtMediaType.XLS_MEDIA_TYPE)
                             .resultMatcher(content().contentType(UniProtMediaType.XLS_MEDIA_TYPE))
                             .build())
+                    .contentTypeParam(ContentTypeParam.builder()
+                            .contentType(UniProtMediaType.OBO_MEDIA_TYPE)
+                            .resultMatcher(content().contentType(UniProtMediaType.OBO_MEDIA_TYPE))
```
LeonardoGonzales: add few String match, to double check the output?

PR #21, opened 2019-07-19
Proteins with and length range facets

LeonardoGonzales:


common-rest/src/main/java/uk/ac/ebi/uniprot/api/common/repository/search/facet/FacetConfigConverter.java
```diff
@@ -16,7 +18,6 @@
  *
  * @author lgonzales
  */
-// TODO: 08/11/18 test
 public interface FacetConfigConverter extends Converter<QueryResponse, List<Facet>> {
 
     /**
```
sahmad-ebi: try to put FacetConfigConverter and GenericFacetConfig common coide in a single abstract class.

PR #27, opened 2019-09-09
Improve search sort order

eddturner:


common-rest/src/main/java/org/uniprot/api/common/repository/search/QueryBoostsFileReader.java
```diff
@@ -0,0 +1,97 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Created 04/09/19
+ *
+ * @author Edd
+ */
+@Slf4j
+public class QueryBoostsFileReader {
```
LeonardoGonzales: Do you think we could use spring configuration logic, like we do with facet? I think each type would be an attribute in the @Configuration mapped class.

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/repository/search/impl/UniProtQueryBoostsConfig.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Created 05/09/19
+ *
+ * @author Edd
+ */
+@Configuration
+public class UniProtQueryBoostsConfig {
+    private static final String BOOSTS_RESOURCE_LOCATION = "uniprotkb-query-boosts.config";
+
```
LeonardoGonzales: Just following the comment above, here we would add the configurations beans, in the same structure that we have in the configuration file..

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
```diff
@@ -130,9 +127,9 @@ private SolrRequest createSolrRequest(SearchRequestDTO request, boolean includeF
         return requestBuilder.build();
     }
 
-    private Sort getUniProtSort(String sortStr, boolean hasScore) {
+    private Sort getUniProtSort(String sortStr) {
         if (Strings.isNullOrEmpty(sortStr)) {
-            return UniProtSortUtil.createDefaultSort(hasScore);
+            return UniProtSortUtil.createDefaultSort(true);
```
LeonardoGonzales: It will always be true? is there any other place that it might be false?  if that is the case, we can remove this parameter. I am not sure....

PR #29, opened 2019-09-25
- Fix Integration tests

LeonardoGonzales:


jieluo321:
It is fine, basically it is refactoring after changing in DataStoreManger.

PR #32, opened 2019-10-01
Trm 22952 repos to read from file not url

eddturner:
I also removed a test class that didn't test anything.

This change speeds up the REST app starting time from 6 minutes -> 6 seconds


uniprotkb-rest/src/test/resources/application.properties
```diff
@@ -44,5 +44,9 @@ streamer.uniprot.batchSize=1000
 streamer.uniprot.valueId=accession_id
 streamer.uniprot.defaultsField=avro_bin
 
+solr.viewby.uniprotCollection=uniprot
+solr.viewby.ecDir=/my/ec/dir
+solr.viewby.uniPathWayFile=/my/unipathway/file
+
```
LeonardoGonzales: missing solr.viewby.keywordFile=

Have you created tests for it? do you think it is worst create it?

eddturner: Great spot.

Yeah I was thinking about tests and I originally wrote some tests, but I ended up testing whether Spring's @ConfigurationProperties was working :) which seemed a bit wrong. So, I think we should just assume that library just works. Are you happy with this?

PR #33, opened 2019-10-02
Trm 22981 investigate slow startup supporting data

eddturner:


eddturner:
There's nothing to change in the core module this way, just how we use it. So it should be fine.

PR #36, opened 2019-10-14
Trm 22591 investigate download problems

eddturner:
This pull request uses cursorMark for the downloads (see https://github.com/ebi-uniprot/uniprot-rest-api/pull/35), in order to preserve ranking of results -- as confirmed by Erick Erickson (https://lucidworks.com/company/), in a thread I had, here:
http://mail-archives.apache.org/mod_mbox/lucene-solr-user/201910.mbox/%3cD9888177-FEA0-4436-AB52-73B887F5E211@gmail.com%3e

This pull request also simplifies `SolrRequestConverter` to only transform to `SolrQuery` instances, and not Spring's `Query` -- to reduce our logic and future maintenance.

*Note* Download integration testing for UniProtKB will be done in a separate Jira, which creates generic download tests for all controllers, using the existing parameterised framework.


uniref-rest/src/main/resources/application.properties
```diff
@@ -36,8 +36,9 @@ voldemort.unref.storeName=uniref
 
 
 ############################### Entry Store (Voldemort) Streaming properties ###############################
-streamer.uniref.batchSize=100
-streamer.uniref.valueId=id
-#streamer.uniref.defaultsField=avro_bin
+streamer.uniref.searchBatchSize=5
```
LeonardoGonzales: is this batch size correct?

uniref-rest/src/main/java/org/uniprot/api/uniref/service/UniRefQueryService.java
```diff
@@ -1,9 +1,5 @@
 package org.uniprot.api.uniref.service;
```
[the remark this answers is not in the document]
eddturner: Yes, let's do it.

uniref-rest/src/main/resources/application.properties
eddturner: i've increased it now.

but the main way to find out the correct size is by doing multiple runs of downloading and adjusting to get best performance.

PR #42, opened 2019-11-22
TRM-23213 Use DeferredResult instead of  responsebodyemitter  for download apis

sahmad-ebi:
While i am testing my changes, can you please take a look at the changes. The change works for all the download apis.


common-rest/src/main/java/org/uniprot/api/rest/controller/BasicSearchController.java
```diff
@@ -114,26 +115,35 @@ protected BasicSearchController(
...
+
+        deferredResult.onTimeout(() -> log.error("Request timeout occurred."));
+
         downloadTaskExecutor.execute(
                 () -> {
                     try {
-                        emitter.send(context, context.getContentType());
-                    } catch (IOException e) {
-                        emitter.completeWithError(e);
+                        deferredResult.setResult(
```
eddturner: Just thinking, do you test what happens when a client does CTRL-C, to abort a download? It'd be good to test to make sure the deferred result handles it gracefully and doesn't wait / consume resources.

common-rest/src/main/java/org/uniprot/api/rest/service/RDFService.java
```diff
@@ -36,7 +40,7 @@ public RDFService(RestTemplate restTemplate, Class<T> clazz) {
     public List<T> getEntries(Iterable<String> accessions) {
         List<String> allAccessions = new ArrayList<>();
         accessions.forEach(acc -> allAccessions.add(acc));
-        LOGGER.debug("RDF call for accessions : {}", allAccessions);
+        LOGGER.info("RDF call for accessions : {}", allAccessions);
```
eddturner: -> debug

common-rest/src/main/java/org/uniprot/api/rest/controller/BasicSearchController.java
```diff
@@ -114,26 +115,35 @@ protected BasicSearchController(
...
+                                ResponseEntity.ok()
+                                        .headers(createHttpDownloadHeader(context, request))
+                                        .body(context));
+                    } catch (Exception e) {
+                        log.error("Error occurred during processing.");
+                        deferredResult.setErrorResult(e);
```
eddturner: Wondering whether we should push this logic down into the service layer, to keep the controller as thin as possible.

eddturner:
PS. I'd also rename the pull request to say it's removing the responsebodyemitter, and replacing with deferredresult (currently it says "TRM-23213 add a custom responsebodyemitter and handler for download api", which is no longer true).

common-rest/src/main/java/org/uniprot/api/rest/controller/BasicSearchController.java
sahmad-ebi: It closes gracefully. The subsequent requests work fine. I need to check if it still keeps some resources open

common-rest/src/main/java/org/uniprot/api/rest/controller/BasicSearchController.java
sahmad-ebi: I agree with this. But it will require a lot of changes. We depend on many fields of BaseController which comes from controller. Can we do this refactoring as a part of another ticket?

LeonardoGonzales: I think this logic is related with how we return the data, and handle any error during it, which,  I think it is controller related logic. Please, correct me if I am wrong.
I think that in the service layer we would put more Business Logic.

proteome-rest/src/main/resources/application.properties
```diff
@@ -34,7 +34,7 @@ spring.data.solr.connectionTimeout=20000
 spring.data.solr.socketTimeout=3600000
 
 # Solr properties
-solr.collectionName=genecentric
+solr.collectionName=proteome
```
LeonardoGonzales: Do you know where do we use this? because you changed from genecentric to proteome, have you noticed anything?

sahmad-ebi: Nope, i didnt notice any issue because of this change.

jieluo321:
now it makes more sense. 
Also i think we should rename the end point to stream rather than download. From api point of view, it is a streaming. Our website uses the stream end point for download.

PR #44, opened 2019-11-28
TRM-23198 - Add framework to test download api and tests for disease download

sahmad-ebi:


common-rest/src/main/java/org/uniprot/api/common/repository/search/SolrResultsIterator.java
```diff
@@ -88,6 +88,10 @@ public void close() {
     private void loadMoreResults() {
         try {
             this.query.set(CursorMarkParams.CURSOR_MARK_PARAM, currentCursorMark);
+            log.info(
+                    "########################### Calling solr server with query {} {}",
```
sahmad-ebi: I will remove it. :-)

PR #52, opened 2020-01-23
Use new searchfields throughout

eddturner:


proteome-rest/src/main/java/org/uniprot/api/proteome/service/GeneCentricSortClause.java
```diff
@@ -29,23 +33,24 @@ protected Sort createDefaultSort(boolean hasScore) {
...
                                 val.getLeft()
                                         .equals(
-                                                GeneCentricField.Sort.accession_id
-                                                        .getSolrFieldName()))) {
+                                                UniProtSearchFields.GENECENTRIC
+                                                        .getSortFieldFor(ACCESSION_ID)
+                                                        .getName()))) {
             return fieldSortPairs;
         } else {
             List<Pair<String, Sort.Direction>> newFieldSortPairs = new ArrayList<>();
```
sahmad-ebi: The problem with using Pair is that it is not very obvious what is in the left and what is in the right. No need to make any change. :-)

proteome-rest/src/main/java/org/uniprot/api/proteome/service/ProteomeSortClause.java
```diff
@@ -8,24 +8,30 @@
...
-import org.uniprot.store.search.field.ProteomeField;
+import org.uniprot.store.search.field.UniProtSearchFields;
 
 /**
  * @author jluo
  * @date: 29 Apr 2019
  */
 @Component
 public class ProteomeSortClause extends AbstractSolrSortClause {
+    private static final String UPID = "upid";
```
sahmad-ebi: We should put such string constant in a util class.

proteome-rest/src/main/java/org/uniprot/api/proteome/service/GeneCentricSortClause.java
```diff
@@ -8,17 +8,21 @@
...
-import org.uniprot.store.search.field.GeneCentricField;
+import org.uniprot.store.search.field.UniProtSearchFields;
 
 /**
  * @author jluo
  * @date: 17 May 2019
  */
 @Component
 public class GeneCentricSortClause extends AbstractSolrSortClause {
+    private static final String ACCESSION_ID = "accession_id";
```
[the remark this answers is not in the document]
eddturner: looking at this `*SortClause` code, to me it looks like all implementations of `AbstractSolrSortClause` are actually doing a similar thing, and probably should be doing the same thing. Maybe we should raise a separate ticket to refactor the sort clause code, and doing so I think this constant would go away.

proteome-rest/src/main/java/org/uniprot/api/proteome/service/GeneCentricSortClause.java
eddturner: Agreed. I would think we should raise a ticket for this and put it on the backlog

PR #58, opened 2020-02-03
Refactor services to use boosting framework used by UniProtKB

eddturner:
UniProtKB boosting framework (edismax + external config defining boosts) has is more configurable and has given users better search results. The goal of this PR is to ensure all appropriate REST services use the same framework.


common-rest/src/main/java/org/uniprot/api/rest/service/BasicSearchService.java
```diff
@@ -161,19 +160,15 @@ protected SolrRequest createSolrRequest(SearchRequest request, boolean includeFa
...
-            DefaultSearchHandler defaultSearchHandler,
-            boolean includeFacets) {
+            boolean includeFacets,
+            QueryBoosts queryBoosts) {
 
         SolrRequest.SolrRequestBuilder requestBuilder = SolrRequest.builder();
 
         String requestedQuery = request.getQuery();
 
         boolean hasScore = false;
```
LeonardoGonzales: I think we do not use hasScore in this method, right? can we remove this line?

support-data-rest/src/main/java/org/uniprot/api/disease/DiseaseService.java
```diff
@@ -1,34 +1,27 @@
...
 import org.springframework.stereotype.Service;
+import org.uniprot.api.common.repository.search.QueryBoosts;
 import org.uniprot.api.rest.service.BasicSearchService;
 import org.uniprot.core.cv.disease.Disease;
-import org.uniprot.store.search.DefaultSearchHandler;
 import org.uniprot.store.search.document.disease.DiseaseDocument;
 import org.uniprot.store.search.field.UniProtSearchFields;
 
 @Service
+@Import(DiseaseQueryBoostsConfig.class)
```
LeonardoGonzales: Do we need to import it?

eddturner: What are your thoughts? For annotation-based configs, then yes, @Import is a normal way of using them.

common-rest/src/main/java/org/uniprot/api/rest/service/BasicSearchService.java
eddturner: it's used a few lines below:

```java
        if (solrSortClause != null) {
            requestBuilder.addSort(solrSortClause.getSort(request.getSort(), hasScore));
        }
```

PR #70, opened 2020-02-27
Write and use the new search config framework

sahmad-ebi:


common-rest/src/test/java/org/uniprot/api/rest/controller/AbstractSearchControllerIT.java
```diff
@@ -391,6 +394,7 @@ void searchCanSearchWithAllAvailableSortFields() throws Exception {
         assertThat(sortFields, not(emptyIterable()));
 
         for (String sortField : sortFields) {
+            System.out.println("====================>" + sortField);
```
eddturner: can remove

support-data-rest/src/test/java/org/uniprot/api/configure/uniprot/domain/AdvanceUniProtKBSearchTermIT.java
```diff
@@ -0,0 +1,106 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.uniprot.api.configure.uniprot.domain.model.AdvanceUniProtKBSearchTerm;
+
+import edu.emory.mathcs.backport.java.util.Arrays;
+
+public class AdvanceUniProtKBSearchTermIT {
```
eddturner: nice

support-data-rest/src/main/java/org/uniprot/api/support_data/controller/UniProtConfigureController.java
```diff
@@ -23,15 +22,10 @@ public UniProtConfigureController(UniProtConfigureService service) {
     }
 
     @GetMapping("/search_terms")
-    public List<SearchItem> getUniProtSearchTerms() {
+    public List<AdvanceUniProtKBSearchTerm> getUniProtSearchTerms() {
```
eddturner: Does this return the new model to front-end? If so, are they okay with this or could it break things for them (if there are minor changes)?

support-data-rest/src/main/java/org/uniprot/api/configure/uniprot/domain/model/AdvanceUniProtKBSearchTerm.java
```diff
@@ -0,0 +1,148 @@
...
+    private List<EvidenceGroup> evidenceGroups;
+
+    @Data
+    @AllArgsConstructor
+    public static class Value {
+        private String name;
+        private String value;
+    }
+
+    public static List<AdvanceUniProtKBSearchTerm> getUniProtKBSearchTerms() {
```
eddturner: This method is called whenever someone goes to the /search_terms in the configure controller. Does it mean multiple calls to `getUniProtKBSearchTerms` will cause repeated navigation of the object structure? If so, can we store the return result in a field, then return it if non-null next time? (badly explained, but hopefully this is clearer:)

```java
private static List<AdvanceUniProtKBSearchTerm> rootSearchTermConfigs;
public static List<AdvanceUniProtKBSearchTerm> getUniProtKBSearchTerms() {
    if(rootSearchTermConfigs == null) {
        // set rootSearchTermConfigs
    }
    return rootSearchTermConfigs;  
}
```

support-data-rest/src/main/java/org/uniprot/api/support_data/controller/UniProtConfigureController.java
sahmad-ebi: I have added old API with static data

PR #75, opened 2020-03-09
Refactor abstract solr clause duplicate reported by sonar

sahmad-ebi:


common-rest/src/main/java/org/uniprot/api/rest/search/AbstractSolrSortClause.java
```diff
@@ -18,34 +20,37 @@
...
+    public Sort getSort(String sortClause) {
+        return StringUtils.isEmpty(sortClause) ? createDefaultSort() : createSort(sortClause);
     }
 
-    public Sort createSort(String sortClause) {
-        return convertToSolrSort(parseSortClause(sortClause));
-    }
-
-    protected abstract Sort createDefaultSort(boolean hasScore);
+    protected abstract List<Pair<String, Sort.Direction>> getDefaultFieldSortOrderPairs();
```
eddturner: It looks like all the implementations of this have more or less the same structure:
```java
        if (this.defaultFieldSortOrderPairs == null) {
            this.defaultFieldSortOrderPairs = new ArrayList<>();
            // add sorts to to this.defaultFieldSortOrderPairs
        }
        return this.defaultFieldSortOrderPairs;
```

If so, you could:
- initialise defaultFieldSortOrderPairs in AbstractSolrSortClause as a new ArrayList<>()
- remove the abstract defaultFieldSortOrderPairs method
- add a concrete method addDefaultFieldOrderPair(field, direction)
- in the implementation classes, call addDefaultFieldOrderPair in initialisation, to assert what the default fields are.

This would then extract only the differing code to the concrete classes, and would save more duplicated code.

Have I missed something? Seems like it'd work to me.

PR #80, opened 2020-03-27
Result field filter json converter

LeonardoGonzales:


common-rest/src/main/java/org/uniprot/api/rest/output/converter/JsonMessageConverter.java
```diff
@@ -78,7 +85,32 @@ protected void before(MessageConverterContext<T> context, OutputStream outputStr
     @Override
     protected void writeEntity(T entity, OutputStream outputStream) throws IOException {
         JsonGenerator generator = tlJsonGenerator.get();
-        generator.writeObject(projectEntryFields(entity));
+        List<ReturnField> fields = getThreadLocalFilterMap();
+        if (Utils.notNullNotEmpty(fields)) {
+            ObjectMapper mapper = objectMapper.copy();
```
eddturner: do we need to do a copy? jackson recommends to reuse objectmappers because they're threadsafe (also see this quick google on stackoverflow https://stackoverflow.com/questions/3907929/should-i-declare-jacksons-objectmapper-as-a-static-field)

common-rest/src/main/java/org/uniprot/api/rest/output/converter/JsonMessageConverter.java
```diff
@@ -78,7 +85,32 @@ protected void before(MessageConverterContext<T> context, OutputStream outputStr
     @Override
     protected void writeEntity(T entity, OutputStream outputStream) throws IOException {
```
eddturner: I think it might be interesting to see what jvisualvm, or java mission control says when writing a big result set? Usually I wouldn't suggest this, but since this method is going to be called a lot, any performance improvements here could be useful.

I wouldn't spend much time doing this, but just to see whether anything stands out and needs optimising a bit.

common-rest/src/main/java/org/uniprot/api/rest/output/converter/JsonMessageConverter.java
LeonardoGonzales: I am manipulating the object mapper and it will have different filters, depending on each request. I am afraid it will keep the previous request behaviour, this is why I copied. I am implementing a test for it. will updated this comment when get the result

PR #82, opened 2020-03-31
update label for publication page

LeonardoGonzales:
Code Change for TRM-23605 and TRM-23606 that are related


uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/PublicationFacetConfig.java
```diff
@@ -81,29 +87,41 @@ static void applyFacetFilters(List<PublicationEntry> publications, PublicationRe
...
+        boolean isLargeScale = value.equalsIgnoreCase(getCleanFacetValue(LARGE_SCALE));
+        return entry.isLargeScale() != isLargeScale;
+    }
+
+    private static boolean filterCategoryValues(String value, PublicationEntry entry) {
+        return entry.getCategories().stream()
+                .noneMatch(
+                        cat -> {
+                            String cleanCategory = getCleanFacetValue(cat);
+                            return cleanCategory.equalsIgnoreCase(value.trim());
```
sahmad-ebi: Optional - It is good to wrap this in a method and use cat -> Predicate

PR #85, opened 2020-04-02
Trm 23752 add dynamic fields comment feature count

sahmad-ebi:
Change in test cases due to two new fields.
@eddturner You may review my last commit only.


support-data-rest/src/test/java/org/uniprot/api/configure/controller/UniProtConfigureControllerIT.java
```diff
@@ -109,6 +109,19 @@ void canGetResultFields() throws Exception {
         validateResponse(response);
     }
 
+    @Test
+    void canGetResultFields2() throws Exception {
```
LeonardoGonzales, line 113: 2? kkkk

sahmad-ebi: renamed to canGetResultFieldsByNewAPI

PR #89, opened 2020-04-16
TRM-23526 use flat search/return fields files in open spec swagger api generation

sahmad-ebi:


eddturner:
ooo, just saw there were a few commented out lines in UniProtKBEntryController -- could u remove them? then all good

PR #91, opened 2020-04-30
Fix content type errors

eddturner:
This PR fixes problems we had before when handling requests like:
```
/accession/P12345.txt -> /accession/P12345 + "Accept: text/flatfile"
/search?query=protein&format=txt -> /search?query=protein + "Accept: text/flatfile"
```

In particular, this PR:
* checks the requested format/extension is valid for the particular endpoint
* that we can handle future endpoints that look like extensions but are not extensions, e.g., maybe we have an endpoint like, /coordinates/P00001/999.1. The logic written ensures it is not treated as an extension.
* errors are handled appropriately.

The class that does the main logic is `HttpServletRequestContentTypeMutator` and the main test class that shows behaviour is, `HttpServletRequestContentTypeMutatorIT`.


jluo321:
We need to document the extension and format behaviour of the request

jluo321:
what extensions and formats we can use, and corresponding response

jluo321:
One more thing is that since MutableHttpServletRequest is based on UniProtKB request and response. I wonder we put HttpCommonHeaderConfig (which uses MutableHttpServletRequest) will have side-effect when this HttpCommonHeaderConfig configuration is applied to other REST API since REST APIs also called this configuraiton

eddturner:
> One more thing is that since MutableHttpServletRequest is based on UniProtKB request and response. I wonder we put HttpCommonHeaderConfig (which uses MutableHttpServletRequest) will have side-effect when this HttpCommonHeaderConfig configuration is applied to other REST API since REST APIs also called this configuraiton

I put `MutableHttpServletRequest` inside common-rest module, and it is invoked by any REST API that uses `HttpCommonHeaderConfig`, so the behaviour it provides comes for free for all the services we have.

eddturner:
> what extensions and formats we can use, and corresponding response

The extensions it accepts are the ones that are available in the end-point. I used some Spring classes to dynamically tell me which formats are valid for each request.

eddturner:
> We need to document the extension and format behaviour of the request

Yes -- will need to do this somewhere?! I'll create a jira to add it somehow to the API documentation.

PR #93, opened 2020-05-14
- add remove() to ThreadLocal

LeonardoGonzales:
add remove() in cleanup method to ThreadLocal variables https://rules.sonarsource.com/java/RSPEC-5164

### DIFF
--- a/common-rest/src/main/java/org/uniprot/api/rest/output/converter/JsonMessageConverter.java
+++ b/common-rest/src/main/java/org/uniprot/api/rest/output/converter/JsonMessageConverter.java
@@ -150,7 +150,7 @@ public class JsonMessageConverter<T> extends AbstractEntityHttpMessageConverter<
         try {
             TL_JSON_GENERATOR.get().close();
         } catch (IOException e) {
-            logger.warn("Unable to close json geneator", e);
+            logger.warn("Unable to close json generator", e);
         }
         TL_JSON_GENERATOR.remove();
         TL_FILTER_MAPPER.remove();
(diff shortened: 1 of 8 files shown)


common-rest/src/main/java/org/uniprot/api/rest/output/converter/JsonMessageConverter.java
```diff
@@ -127,21 +128,33 @@ protected void writeEntity(T entity, OutputStream outputStream) throws IOExcepti
...
+    }
+
+    @Override
+    protected void cleanUp() {
+        super.cleanUp();
+        try {
+            TL_JSON_GENERATOR.get().close();
+        } catch (IOException e) {
+            logger.warn("Unable to close json geneator", e);
+        }
```
sahmad-ebi, line 154: Typo. `generator`

common-rest/src/main/java/org/uniprot/api/rest/output/converter/XlsMessageConverter.java
```diff
@@ -44,49 +44,53 @@ public XlsMessageConverter(
...
-        tlWorkBook.get().dispose();
+        TL_WORK_BOOK.get().write(outputStream);
+        TL_WORK_BOOK.get().dispose();
     }
 
     @Override
     protected void cleanUp() {
         try {
-            tlWorkBook.get().close();
+            TL_WORK_BOOK.get().close();
```
sahmad-ebi, line 67: We should call super.cleanUp();

PR #98, opened 2020-05-20
remove all uses of spring data solr

eddturner:
... and refactor to directly use apache solr's solrj library.


eddturner:
Yes it's really weird that I couldn't find the Operator type in Solrj ... maybe I'm blind -- I looked, and google'd, but didn't find anything.

Anyway, thank you for reviewing!

PR #102, opened 2020-06-01
proteome and genecentric swagger document

jluo321:
proteome and gene centric swagger document, gene centric can only see two endpoints

### DIFF
--- a/support-data-rest/pom.xml
+++ b/support-data-rest/pom.xml
@@ -12,16 +12,11 @@
 	<artifactId>support-data-rest</artifactId>
 	<description>UniProt Rest API for support data</description>
 
-	<properties>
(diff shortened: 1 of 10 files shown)


support-data-rest/pom.xml
```diff
@@ -12,16 +12,11 @@
...
-	<properties>
-		<dropwizard.metrics.version>3.2.6</dropwizard.metrics.version>
-		<openapi-maven-plugin.version>0.3</openapi-maven-plugin.version>
-	</properties>
 
 	<dependencies>
 		<dependency>
 			<groupId>io.dropwizard.metrics</groupId>
 			<artifactId>metrics-core</artifactId>
-			<version>${dropwizard.metrics.version}</version>
```
sahmad-ebi, line 24: Why did you remove the version of dropwizard?

PR #104, opened 2020-06-11
add query optimisation for default search

LeonardoGonzales:


jieluo321:
Looks good. I noticed that for uniref, you added two fields for optimising (id, upi). For uniprotkb, we could also add more optimised fields in the future to optimise the query (one possible example is EC number).

 I wonder if it will generate some syntax error in some edge cases (complex query).

common-rest/src/main/java/org/uniprot/api/rest/service/DefaultSearchQueryOptimiser.java
```diff
@@ -0,0 +1,105 @@
...
+public class DefaultSearchQueryOptimiser {
+
+    private final List<SearchFieldItem> optimisedFields;
+
+    public DefaultSearchQueryOptimiser(SearchFieldItem idField) {
+        this(Collections.singletonList(idField));
+    }
+
+    public DefaultSearchQueryOptimiser(List<SearchFieldItem> optimisedFields) {
+        this.optimisedFields = optimisedFields;
```
eddturner, line 34: call it `optimisableFields` or `fieldsToOptimise`, so it doesn't sound like these fields are already optimised?

common-rest/src/main/java/org/uniprot/api/rest/service/DefaultSearchQueryOptimiser.java
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+        } catch (ParseException e) {
+            throw new ServiceException(
+                    "Unable to parse query in the optimisation" + requestedQuery, e);
+        }
+        return result;
+    }
+
+    private List<String> getDefaultSearchFieldTerms(Query query) {
```
eddturner, line 66: Would it make sense to recursively iterate through `query` (as you already do), then just return the `List<TermQuery>` (call it `X`) whose field names are `""`. Then, `getOptimisedDefaultTermForValue` can mutate the element `e` of `X` if the `e` matches any field in `optimisableFields`'s regexes?

This way we don't do anything special with "" and the query structure isn't being changed -- we're only adding field names where applicable. Maybe this way seems less scary?

PR #107, opened 2020-06-23
Limit streaming

eddturner:


common-rest/src/main/java/org/uniprot/api/common/repository/store/TupleStreamTemplate.java
```diff
@@ -34,12 +40,17 @@
 public class TupleStreamTemplate {
     private final StreamerConfigProperties streamConfig;
     private final HttpClient httpClient;
+    private final SolrClient solrClient;
+    private final SolrCollection collection;
```
LeonardoGonzales, line 44: we already have collection in the streamConfig attribute that we have in line 41

common-rest/src/main/java/org/uniprot/api/common/repository/store/StreamerConfigProperties.java
```diff
@@ -14,6 +14,7 @@
     private int storeBatchSize;
     private int storeFetchMaxRetries;
     private int storeFetchRetryDelayMillis;
+    private int storeMaxCountToRetrieve;
 
     private String zkHost;
     private String idFieldName;
```
LeonardoGonzales, line 20: Hi Edd do you think we could reuse Solr collection that we have in this class in the TupleStreamTemplate validation method? I pointed out bellow in TupleStreamTemplate

common-rest/src/main/java/org/uniprot/api/common/repository/store/TupleStreamTemplate.java
eddturner: woops thanks!

PR #108, opened 2020-06-24
Trm 24148 get by accessions

ahmadshadab:
Summary of behaviour of the API "uniprotkb/accessions".

1. accessions param is mandatory and it can take upto 1000 comma separated accessions.
2. If size is not passed, all the entries for the passed accessions will be returned in the response.
3. If size is passed and its value is less than number of accessions then a cursor will be returned in the header to request next page.
4. If size is passed and its value is more than number of accessions then all entries will be returned in a single page.
5. facets param is optional and can take comma separated list of valid facet names.
6. facets are returned only in the first page response of a paginated request. facets won't be returned from 2nd page onwards even if you pass the parameter with valid values.
7. To start download in a new file, pass download=true.(By default it is false)
8. facets param is ignored when download=true.
9. all content types are supported.
10. facetFilter takes any valid solr query constructed using only facet field(s).


jieluo321:
about AccessionListValidator, if all the accessions except for one have right syntax, it will fail the validation. I don't know if this will annoy people. (one accession endpoint makes sense, but i wonder we need so restrict validator for a list of accessions)

if the number of accessions is more than maximum number of size it will fail. I also don't know about it. i wonder if we only process supported maximum number of accession

jieluo321:
code coverage for StoreStreamer and TupleStreamTemplate can be improved

jieluo321:
ValidAccessionList, do we want to have this behaviour? if size of list accessions is larger than maximum size, validation failure
If any of accessions does not meet accession regular expression, validation failure. I am not sure we need so restrict validation. 

Also ValidAccessionListTest test accession is in expression of "P12345, P54321, P32542". I think that our query strings are without space only comma separated.

jieluo321:
final comment is facet filter. I think that Shadab and Leo are addressing this issue

PR #113, opened 2020-07-17
Uniref light

LeonardoGonzales:


common-rest/src/main/java/org/uniprot/api/common/repository/search/QueryResult.java
```diff
@@ -1,6 +1,7 @@
 package org.uniprot.api.common.repository.search;
```
ahmadshadab, line 1: Should we rename the controller test class with prefix UnirefLight. e.g. UniRefLightSearchControllerIT and UniRefLightStreamControllerIT

PR #114, opened 2020-07-20
TRM-24510 Improve postman test framework by adding verify list of acc…

ahmadshadab:
…essions


jieluo321:
For the time being, it is ok. We may add more verify script as we start to add more test cases to uniprotkb.

PR #115, opened 2020-07-28
accept compressed request parameter and mutate it to accept encoding …

LeonardoGonzales:
Before this code change, we would only identify the compressed(gzip) option through "Accept-Encoding" request header.
But front end is unable to send us request headers for the download use case.
so, now we also have to accept request parameter compressed=true. 
The way the we implemented it to make it flexible is during the request filter we identify it and add  "Accept-Encoding" to request header when we have a request parameter compressed=true....

### DIFF
--- a/common-rest/src/main/java/org/uniprot/api/rest/request/HttpServletRequestContentTypeMutator.java
+++ b/common-rest/src/main/java/org/uniprot/api/rest/request/HttpServletRequestContentTypeMutator.java
@@ -107,7 +107,7 @@ public class HttpServletRequestContentTypeMutator {
 
     private void mutateAcceptEncodingHeader(MutableHttpServletRequest request) {
         String compressed = request.getParameter(COMPRESSED);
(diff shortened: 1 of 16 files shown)


common-rest/src/main/java/org/uniprot/api/rest/request/HttpServletRequestContentTypeMutator.java
```diff
@@ -96,6 +105,13 @@ public void mutate(
         }
     }
 
+    private void mutateAcceptEncodingHeader(MutableHttpServletRequest request) {
+        String compressed = request.getParameter(COMPRESSED);
+        if (Utils.notNullNotEmpty(compressed) && compressed.equalsIgnoreCase("true")) {
+            request.addHeader(HttpHeaders.ACCEPT_ENCODING, FileType.GZIP.getFileType());
```
ahmadshadab, line 111: One condition is sufficient here 
`"true".equalsIgnoreCase(compressed)`

PR #118, opened 2020-08-28
Uniref edismax qf

eddturner:
updates common-rest, to allow parsing of qf values in config file, to be used when generating queries, becoming part of the qf parameter sent to Solr


common-rest/src/main/java/org/uniprot/api/common/repository/search/SolrQueryConfigFileReader.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private void initialiseSolrQueryConfig() {
+        log.info("Loading Solr query config [" + this.resourceLocation + "]...");
+        InputStream resourceAsStream = getClass().getResourceAsStream(resourceLocation);
+        if (notNull(resourceAsStream)) {
+
+            Stream<String> lines =
+                    new BufferedReader(new InputStreamReader(resourceAsStream)).lines();
+            QueryConfigType queryConfigType = QueryConfigType.DEFAULT_SEARCH;
+            for (String line : lines.toArray(String[]::new)) {
```
LeonardoGonzales, line 41: Edd, this is a suggestion/discussion, but, do you think we could use switch case here? 
because this prefix is the entire line, right? 
We would have to change from enum to constants do you think that it is worst?
What do you prefer?

PR #120, opened 2020-09-02
Add entry page facet and update member id type enum and convert xml b…

LeonardoGonzales:
Add entry page facet and update member id type enum and convert xml behaviour

### DIFF
--- a/uniref-rest/src/main/java/org/uniprot/api/uniref/repository/store/UniRefEntryFacetConfig.java
+++ b/uniref-rest/src/main/java/org/uniprot/api/uniref/repository/store/UniRefEntryFacetConfig.java
@@ -1,0 +1,158 @@
+package org.uniprot.api.uniref.repository.store;
+
+import static org.uniprot.api.common.repository.search.facet.FacetUtils.getCleanFacetValue;
+import static org.uniprot.store.search.SolrQueryUtil.*;
+
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import org.springframework.stereotype.Component;
+import org.uniprot.api.common.repository.search.facet.*;
+import org.uniprot.core.uniref.UniRefEntryLight;
+import org.uniprot.core.uniref.UniRefMemberIdType;
+import org.uniprot.core.util.Utils;
+
+/**
+ * @author lgonzales
+ * @since 26/08/2020
+ */
+@Component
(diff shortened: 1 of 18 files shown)


uniref-rest/src/main/java/org/uniprot/api/uniref/repository/store/UniRefEntryFacetConfig.java
```diff
@@ -0,0 +1,158 @@
...
+                        .values(getUniProtMemberTypeValues(uniRefEntryLight.getMembers()))
+                        .build();
+        if (Utils.notNullNotEmpty(uniProtMemberType.getValues())) {
+            results.add(uniProtMemberType);
+        }
+
+        return results;
+    }
+
+    private static boolean filterMemberTypeValue(String filterValue, String memberValue) {
```
ahmadshadab, line 102: Please put some comment here as well since it is negation of what one can think of.

uniref-rest/src/main/java/org/uniprot/api/uniref/service/UniRefLightSearchService.java
```diff
@@ -73,8 +73,14 @@ public UniRefEntryLight findByUniqueId(String uniqueId, String fields) {
...
+            Stream<UniRefEntryLight> content =
+                    result.getContent().map(this::removeOverLimitAndCleanMemberId);
+
+            result = QueryResult.of(content, result.getPage(), result.getFacets());
+        } else {
+            Stream<UniRefEntryLight> content = result.getContent().map(this::cleanMemberId);
+
+            result = QueryResult.of(content, result.getPage(), result.getFacets());
         }
         return result;
```
ahmadshadab, line 85: we can take out             result = QueryResult.of(content, result.getPage(), result.getFacets());
from if and else blocks.

uniref-rest/src/test/java/org/uniprot/api/uniref/controller/UniRefGetIdControllerIT.java
```diff
@@ -223,7 +234,7 @@ void getFirstPageMembers() throws Exception {
                                         .header(ACCEPT, APPLICATION_JSON_VALUE));
 
         // then
-        response.andDo(log())
+        response.andDo(print())
                 .andExpect(status().is(HttpStatus.OK.value()))
```
ahmadshadab, line 238: change print to log otherwise gitlab might complain

PR #124, opened 2020-09-22
Create Get by taxon ids

LeonardoGonzales:


support-data-rest/src/main/java/org/uniprot/api/support/data/taxonomy/request/GetByTaxonIdsRequest.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * @author lgonzales
+ * @since 17/09/2020
+ */
+@Data
+public class GetByTaxonIdsRequest implements SearchRequest {
+
+    private static final String TAXONOMY_ID_LIST_REGEX = "^\\d+(?:,\\d+)*$";
+
```
ahmadshadab, line 32: Please make this String constant hidden otherwise it will be generated in swagger doc.
@Parameter(hidden = true)

PR #125, opened 2020-09-22
Add uniparc id for rest service

LeonardoGonzales:


ahmadshadab (commented):
Leo, i didn't see change in the rest call to see uniParcId being returned in the response?
Can you please add one if not already added.

PR #126, opened 2020-09-28
White list fields

LeonardoGonzales:
In Our UniProtKB database we have some ids that has the following format <prefix><colon><postfix>,
 for example (HGNC:3689).
 These Ids, are processed wrongly by solr query parser, basically, it identify HGNC as a field name, causing
 an error in the request, because we do not have a defined field named HGNC in our solr schema.

 In order to sort out this problem, we created this property file with a list of trouble ids that will need to be
 whitelisted in our field validation and handled specially in our UniProtFieldQueryNodeProcessor.


eddturner (approved):
great

PR #127, opened 2020-10-01
Add max validation to page size and update its logic

LeonardoGonzales:


common-rest/src/main/resources/common-message.properties
```diff
@@ -1,5 +1,5 @@
 search.required='{field}' is a required parameter
-search.positive.or.zero='{field}' must be greater than or equal to 0
+search.max.page.size='{field}' must be less than or equals to {value}
```
eddturner, line 2: equals -> equal

eddturner (approved):
one typo (commented in review) but other than that, merge it, go for it!

PR #129, opened 2020-10-21
Uniprotkb edismax qf

eddturner:


common-rest/src/test/java/org/uniprot/api/rest/service/query/UniProtQueryProcessorTest.java
```diff
@@ -18,6 +18,7 @@
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.uniprot.api.rest.service.query.processor.UniProtQueryNodeProcessorPipeline;
+import org.uniprot.api.rest.service.query.processor.UniProtQueryProcessorConfig;
```
LeonardoGonzales, line 21: Just as curiosity, what happens if I search (protein and gene) in the search? and this (and) is just a word that is part of the text.

eddturner: the way it's configured is to only treat "AND" as the logical operator. Otherwise, "and" is treated as any normal word.

However, since "and" is actually a stopword, it is in fact removed from the request sent to Solr.

The test-cases to look out for are these ones:
```java
@Test
    void ignoreTheAndStopWordThenInterpretDefaultQuery() {
        String processedQuery = processor.processQuery("a and b");
        // -> remove "and" stopword from query
        // -> "a b" 
        // -> "a AND b" (because AND is default operator)
        assertThat(processedQuery, is("a AND b"));
    }

    @Test
    void ignoreTheOrStopWordThenInterpretDefaultQuery() {
        String processedQuery = processor.processQuery("a or b");
        // -> remove "or" stopword from query
        // -> "a b" 
        // -> "a AND b" (because AND is default operator)
        assertThat(processedQuery, is("a AND b"));
    }
```

The reason we need to remove stopwords from the query manually is a bit complicated:
* we ignore stopwords during indexing
* edismax nested queries has required us to manually insert AND/ORs between terms, for non-phrase queries. E.g., "hello world" -> "hello AND world"
* if someone sends "hello and world", the simple transformation is "hello AND and AND world"
* the last query above "... AND and ..." means Solr will*require* "and" to be present. But since indexing removed the "and", then it will find nothing
* therefore, we remove the "and" from the query, so that "hello and world" is transformed to "hello AND world".

I hope that makes sense!

PR #131, opened 2020-11-02
GeneCentric changes

LeonardoGonzales:


proteome-rest/src/main/java/org/uniprot/api/proteome/request/GeneCentricStreamRequest.java
```diff
@@ -0,0 +1,28 @@
...
+ * @author lgonzales
+ * @since 29/10/2020
+ */
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class GeneCentricStreamRequest extends GeneCentricBasicRequest implements StreamRequest {
+
+    @Parameter(
+            description =
+                    "Add content disposition attachment to response header, this way it can be downloaded as a file in the browser.")
```
eddturner, line 22: Add -> Adds

eddturner: header -> headers

proteome-rest/src/main/java/org/uniprot/api/proteome/controller/GeneCentricController.java
```diff
@@ -187,18 +182,25 @@ public GeneCentricController(
...
+                LIST_MEDIA_TYPE_VALUE,
+                APPLICATION_JSON_VALUE,
+                FASTA_MEDIA_TYPE_VALUE,
+                APPLICATION_XML_VALUE
+            })
+    public DeferredResult<ResponseEntity<MessageConverterContext<GeneCentricEntry>>> stream(
+            @Valid @ModelAttribute GeneCentricStreamRequest streamRequest,
+            HttpServletRequest request) {
+        Stream<GeneCentricEntry> result = service.stream(streamRequest);
+        return super.download(result, streamRequest, getAcceptHeader(request), request);
```
eddturner, line 203: should we rename super.download -> stream?

eddturner (changes requested):
very minor comments, just 1 question about whether **download** should be called **stream** in basicsearchcontroller

PR #136, opened 2020-12-03
provide rdf/xml format for streaming uniref entries

ahmadshadab:


uniparc-rest/src/main/resources/application.properties
```diff
@@ -23,15 +23,16 @@ serviceInfoPath=classpath:service-info.json
...
 #spring.data.solr.zkHost=edd:2191
 #spring.data.solr.zkHost=localhost:2181
 #spring.data.solr.httphost=http://wp-np2-b4:8983/solr/
 #spring.data.solr.httphost=https://www.ebi.ac.uk/uniprot/services/search/
 #spring.data.solr.httphost=http://localhost:8082/solr/
+spring.data.solr.zkHost=wp-np2-b9.ebi.ac.uk:4191,ves-hx-ee.ebi.ac.uk:4191,ves-hx-c9.ebi.ac.uk:4191
 
-spring.data.solr.username=
-spring.data.solr.password=
+spring.data.solr.username=solr_admin
```
eddturner, line 34: remove username+password

eddturner (changes requested):
Would it be possible to have an additional test to check the _contents_ of a request that asks for RDF content type? Basically, if the mock RestTemplate (used to fetch from the RDF service) is wired up with some dummy data, then we check that the expected data + RDF_PROLOG is returned. This will just show that we've wired things up properly.

PR #137, opened 2020-12-04
update UniRef

LeonardoGonzales:
- add cluster solr search field and enable it in the search
- UniRefEntryLight  model change: commonOrganism id and name moved to Organism 

- add retry logic to datastore and solr index to try to avoid 11PM errors


ahmadshadab (approved):
looks good

PR #142, opened 2021-01-26
Support data tuning

ahmadshadab:
Main changes
- remove content
- use stream
- support obo
- update swagger doc


eddturner (changes requested):
I've whizzed through this and i think it looks good. Quick question: did u update the end-points (for supporting-data) as discussed in the Monday morning meetings (https://www.ebi.ac.uk/seqdb/confluence/pages/viewpage.action?spaceKey=UniProt&title=1+Feb+-+Remote+meeting)?  Maybe they're all fine -- but i'm not 100% sure.

PR #143, opened 2021-01-26
Update UniParc model TRM-25358 and TRM-25359

LeonardoGonzales:


uniparc-rest/src/test/java/org/uniprot/api/uniparc/controller/UniParcControllerGetBySequenceIT.java
```diff
@@ -201,15 +200,14 @@ void testGetBySequenceSuccess() throws Exception {
...
+                .andExpect(jsonPath("uniParcCrossReferences[*].proteomeId", hasItem("UP123456701")))
+                .andExpect(jsonPath("uniParcCrossReferences[*].taxonomy", notNullValue()))
                 .andExpect(
                         jsonPath(
-                                "uniParcCrossReferences[*].properties[*].value",
-                                hasItem("UP123456701")))
+                                "uniParcCrossReferences[*].taxonomy.scientificName",
+                                notNullValue()))
+                .andExpect(jsonPath("uniParcCrossReferences[*].taxonomy.taxonId", notNullValue()))
                 .andExpect(jsonPath("sequence", notNullValue()))
```
ahmadshadab, line 211: Are we also going to return common name of taxonomy?

uniparc-rest/src/main/resources/uniparc.facet.properties
```diff
@@ -3,6 +3,7 @@ facet.limit=5
 
 facet.uniparc.database.label=Database
 facet.uniparc.database.allowmultipleselection=true
+facet.uniparc.database.limit=-1
```
ahmadshadab, line 7: It would be better if we could put a description how this limit=-1 works when we already have facet.limit=5

ahmadshadab: If we could the format `f.<fieldname>.facet.limit` then it will help us in making similar change in other config file in future

LeonardoGonzales: I added it in FacetProperty object. (This is a shared place (common-rest))

PR #149, opened 2021-02-04
rename UniParcCrossReference.taxonomy to UniParcCrossReference.organism

LeonardoGonzales:


uniparc-rest/src/test/java/org/uniprot/api/uniparc/controller/UniParcControllerITUtils.java
```diff
@@ -97,7 +97,7 @@ static UniParcCrossReference getXref(UniParcDatabase database) {
                         .created(LocalDate.of(2017, 2, 12))
                         .lastUpdated(LocalDate.of(2017, 4, 23))
                         .proteinName(getName("anotherProteinName", i))
-                        .taxonomy(getTaxonomy(9606L))
+                        .organism(getTaxonomy(9606L))
                         .proteomeId("UP000005640")
```
ahmadshadab, line 101: better if we rename getTaxonomy to getOrganism

PR #152, opened 2021-02-09
add allDatabases endpoint for UniParcConfigureController

LeonardoGonzales:


ahmadshadab (approved):
Approved.
Is field name uriLink ok or should we use url?

PR #154, opened 2021-02-16
allow size>=0 for uniref members

ahmadshadab:


uniref-rest/src/test/java/org/uniprot/api/uniref/controller/UniRefMembersControllerIT.java
```diff
@@ -393,6 +415,42 @@ void invalidFacetsBadRequest() throws Exception {
...
     }
 
+
+    @Test
+    void getMembersSuccessSizeZero() throws Exception {
+        // when
+        ResultActions response =
+                mockMvc.perform(
+                        get(MEMBER_PREFIX_PATH + ID_100 + MEMBER_SUFIX_PATH)
+                                .param("size", "0")
```
LeonardoGonzales, line 425: The use case for pageSize=0 is with Facets. Can you please add facet parameter here and make sure it is returning with counts. I added a similar test in the AbstractSearchIT, you can check there.

LeonardoGonzales (commented):
Just a minor improvement for test. You can merge it to master after the change.

PR #155, opened 2021-02-17
Uniparc acc column fix and database label

LeonardoGonzales:


ahmadshadab (commented):
Could we add tests to verify the value of the labels due to this change?

PR #158, opened 2021-03-22
Get by ids api

ahmadshadab:


uniref-rest/src/main/java/org/uniprot/api/uniref/request/UniRefIdsSearchRequest.java
```diff
@@ -0,0 +1,81 @@
...
+    private String fields;
+
+    @Parameter(description = "Name of the facet search")
+    @ValidFacets(facetConfig = UniRefFacetConfig.class)
+    @ValidContentTypes(contentTypes = {MediaType.APPLICATION_JSON_VALUE})
+    private String facets;
+
+    @Parameter(
+            description =
+                    "Criteria to filter by facet value. It can any supported valid solr query.")
```
eddturner, line 51: any supported -> support any

find/replace solr for lucene (throughout in comments -- i think we said we'd do this once before for other descriptions, in a conv with leo)

eddturner (approved):
looks good. If you could run it locally against live solr/voldemort, that'd be great. if it all hangs together, then merge

uniref-rest/src/main/java/org/uniprot/api/uniref/request/UniRefIdsSearchRequest.java
ahmadshadab: Thanks.. copy paste error :-)

PR #160, opened 2021-04-07
Support data citation improvement

LeonardoGonzales:


support-data-rest/src/main/java/org/uniprot/api/support/data/literature/controller/LiteratureController.java
```diff
@@ -199,8 +199,8 @@ public LiteratureController(
 
     @Override
     protected String getEntityId(LiteratureEntry entity) {
-        Literature literature = (Literature) entity.getCitation();
-        return String.valueOf(literature.getPubmedId());
+        Citation literature = entity.getCitation();
+        return String.valueOf(literature.getId());
     }
```
ahmadshadab, line 204: We can remove String.valueOf since getId returns int

support-data-rest/src/test/java/org/uniprot/api/support/data/literature/controller/LiteratureITUtils.java
```diff
@@ -59,34 +60,36 @@ public static LiteratureDocument createSolrDoc(long pubMedId, boolean facet) {
...
-
-        return document;
+        List<String> citationsWith = new ArrayList<>();
+        if (facet) {
+            citationsWith.add("1_uniprotkb");
+            citationsWith.add("2_reviewed");
+            citationsWith.add("3_unreviewed");
+            citationsWith.add("4_computationally");
+            citationsWith.add("5_community");
+        }
```
ahmadshadab, line 70: Use global enum

ahmadshadab (commented):
I am sure that some Abstract class test will be searching UniProtKB entry by field lit_citation_id

PR #163, opened 2021-04-15
update behaviour for deleted taxonomy and excluded proteome and also …

LeonardoGonzales:
…add ancestor search for taxonomy


proteome-rest/src/main/java/org/uniprot/api/proteome/service/ProteomeQueryService.java
```diff
@@ -40,6 +44,17 @@ public ProteomeQueryService(
...
+    protected SolrRequest.SolrRequestBuilder createSolrRequestBuilder(
+            BasicRequest request,
+            AbstractSolrSortClause solrSortClause,
+            SolrQueryConfig queryBoosts) {
+        SolrRequest.SolrRequestBuilder builder =
+                super.createSolrRequestBuilder(request, solrSortClause, queryBoosts);
+        builder.filterQuery(getActiveProteomeFilterQuery());
+        return builder;
+    }
+
```
ahmadshadab, line 57: We can do it later. But i feel, the value of argument(EXCLUDED_FIELD + ":false") in method builder.filterQuery should come from BasicRequest object. That way the service will be independent of UI business logic. If someone else can use our service class without filtering by excluded.

support-data-rest/src/main/java/org/uniprot/api/support/data/taxonomy/service/TaxonomyService.java
```diff
@@ -41,6 +47,26 @@ public TaxonomyService(
...
+            AbstractSolrSortClause solrSortClause,
+            SolrQueryConfig queryBoosts) {
+        SolrRequest.SolrRequestBuilder builder =
+                super.createSolrRequestBuilder(request, solrSortClause, queryBoosts);
+
+        if(!(request instanceof GetByTaxonIdsRequest)) {
+            //We do not add active filter query for get by Taxon Ids
+            //Because this endpoint need to return excluded taxon ids
+            builder.filterQuery(getActiveTaxonomyFilterQuery());
+        }
```
ahmadshadab, line 62: Similar comment like ProteomeQueryService

support-data-rest/src/test/java/org/uniprot/api/support/data/taxonomy/controller/TaxonomyGetIdControllerIT.java
```diff
@@ -136,6 +138,27 @@ void validMergedEntryReturnRedirectSeeOther() throws Exception {
...
+
+        // then
+        response.andDo(log())
+                .andExpect(status().is(HttpStatus.OK.value()))
+                .andExpect(
+                        header().string(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE))
+                .andExpect(jsonPath("$.taxonId", is(Integer.parseInt(DELETED_TAX_ID))))
+                .andExpect(jsonPath("$.active", is(false)))
+                .andExpect(jsonPath("$.inactiveReason.inactiveReasonType", is("DELETED")));
+    }
```
ahmadshadab, line 160: we may use enum TaxonomyInactiveReasonType here for verify DELETED value

ahmadshadab (commented):
Nice test coverage as usual. Just few small comments except one

PR #164, opened 2021-04-23
add idmapping job detail endpoint

LeonardoGonzales:


idmapping-rest/src/main/java/org/uniprot/api/idmapping/controller/IdMappingJobController.java
```diff
@@ -97,6 +98,42 @@ public IdMappingJobController(
...
+    @Operation(
+            summary = "Get the details of a job.",
+            responses = {
+                    @ApiResponse(
+                            content = {
+                                    @Content(
+                                            mediaType = APPLICATION_JSON_VALUE,
+                                            array =
+                                            @ArraySchema(
+                                                    schema =
```
ahmadshadab, line 114: This should be single entry response for Impl JobDetailResponse.class

idmapping-rest/src/test/java/org/uniprot/api/idmapping/controller/IdMappingJobControllerIT.java
```diff
@@ -379,4 +382,91 @@ void jobSubmittedMoreThanAllowedIdsFailed() throws Exception {
...
+                        .jobStatus(JobStatus.FINISHED)
+                        .jobId(jobId)
+                        .build();
+        when(cacheService.getJobAsResource(jobId)).thenReturn(job);
+
+        // when
+        ResultActions response =
+                mockMvc.perform(
+                        get(JOB_DETAILS_ENDPOINT + "/" + jobId)
+                                .header(ACCEPT, MediaType.APPLICATION_JSON));
```
ahmadshadab, line 444: Optional - If we use method MockMvcRequestBuilders.get(String urlTemplate, Object... uriVars), it will look cleaner.

ahmadshadab (commented):
Just few minor comments. Please merge once you fix at least Swagger issue.

PR #165, opened 2021-04-26
improve annotation tag and fix evidence level tag

LeonardoGonzales:


common-rest/src/test/java/org/uniprot/api/common/repository/search/facet/FacetResponseConverterTest.java
```diff
@@ -79,6 +79,36 @@ void convertFacetWithoutLabel() {
...
+        assertNull(itemValue.getLabel());
+        assertEquals("5", itemValue.getValue());
+        assertEquals(Long.valueOf(51L), itemValue.getCount());
+
+        itemValue = annotation.getValues().get(1);
+        assertNotNull(itemValue);
+        assertNull(itemValue.getLabel());
+        assertEquals("4", itemValue.getValue());
+        assertEquals(Long.valueOf(41L), itemValue.getCount());
+    }
```
eddturner, line 110: is it possible for there to be two values with the same count? Is it easy to add another section like lines 105-109, with itemValue.getValue() == 4 again, just to nail down that edge case?

eddturner (approved):
Good stuff!

common-rest/src/test/java/org/uniprot/api/common/repository/search/facet/FacetResponseConverterTest.java
LeonardoGonzales: It is not possible because we populate it in a fake mocked config.

PR #168, opened 2021-05-04
add releaseNumber and releaseDate to service info and update header r…

LeonardoGonzales:
…esponse to return both

### DIFF
--- a/common-rest/src/main/java/org/uniprot/api/rest/output/header/HttpCommonHeaderConfig.java
+++ b/common-rest/src/main/java/org/uniprot/api/rest/output/header/HttpCommonHeaderConfig.java
@@ -30,10 +30,10 @@ import org.uniprot.api.rest.service.ServiceInfoConfig;
 @Configuration
 @Import({ServiceInfoConfig.class})
 public class HttpCommonHeaderConfig {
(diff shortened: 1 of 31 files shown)


common-rest/src/main/java/org/uniprot/api/rest/output/header/HttpCommonHeaderConfig.java
```diff
@@ -30,8 +30,10 @@
 @Configuration
 @Import({ServiceInfoConfig.class})
 public class HttpCommonHeaderConfig {
-    public static final String X_RELEASE = "X-Release";
+    public static final String X_RELEASE_NUMBER = "X-ReleaseNumber";
+    public static final String X_RELEASE_DATE = "X-ReleaseDate";
     static final String ALLOW_ALL_ORIGINS = "*";
+    public static final String X_TOTAL_RECORDS = "X-TotalRecords";
```
eddturner, line 36: maybe these should not be camelcase?

I've seen sometimes servers/traffic managers changing the headers to be lowercase (no idea why this happens sometimes).

so maybe they could be
x-release-number
x-release-date
x-total-records ?

eddturner (changes requested):

PR #169, opened 2021-05-05
improve FacetStream to allow descending order values

LeonardoGonzales:
The use case is annotation_score facet. Order from 5 to 1 descending. The fix is already working for search facet. This code change it to add it to stream facet.


ahmadshadab (approved):
Looks good. Did you try integration test with postman/curl with test1/wwwdev solr server?

PR #176, opened 2021-05-14
no caching for idmapping

eddturner:


LeonardoGonzales (approved):
Looks good for me. The only suggestion I have is to add these paths in the application.
properties (if you agree open a Jira for the future)

eddturner:
Thanks @LeonardoGonzales . I refactored it to make the list of non-cacheable paths to be configurable in a properties file now. Thanks!

PR #178, opened 2021-05-20
lower case go evidence code value for uniprotkb advanced search

LeonardoGonzales:


support-data-rest/src/main/java/org/uniprot/api/support/data/configure/response/AdvancedSearchTerm.java
```diff
@@ -135,6 +142,23 @@ private static AdvancedSearchTerm from(SearchFieldItem fi) {
         return b.build();
     }
 
+    private static EvidenceGroup mapGoEvidenceGroup(EvidenceGroup evidenceGroup) {
+        EvidenceGroupImpl result = new EvidenceGroupImpl();
```
ahmadshadab, line 146: Type should be of Interface `EvidenceGroup`

ahmadshadab (approved):
Looks good
just one minor comment.

PR #182, opened 2021-06-29
Facet improve

LeonardoGonzales:


eddturner (changes requested):
Once the ranking issue is fixed, this is approved

PR #194, opened 2021-08-24
add tests for new fields added to unirule and arba

ahmadshadab:


jieluo321 (commented):
There are very few search tests in the rest api. I checked uniprot-indexer-integration-test module in uniprot-store, there are not aa test either. We need to add the aa search test in uniprot-indexer-integration-test module, to be consistent with other indices and searches.

PR #207, opened 2021-10-26
Chebi related ids (is_a)

LeonardoGonzales:


aa-rest/pom.xml
```diff
@@ -6,7 +6,6 @@
 		<groupId>org.uniprot</groupId>
 		<artifactId>uniprot-rest-api</artifactId>
 		<version>1.0.8-SNAPSHOT</version>
-		<relativePath>../</relativePath>
```
eddturner, line 9: removing this means we depend on the installed pom, rather than the pom available in the git repo. i think we should add this back. Same for everywhere the path was removed.

eddturner (changes requested):

PR #212, opened 2021-11-08
UniProtQueryProcessor to be created for every request, to prevent many exceptions currently being seen.

eddturner:
create new UniProtQueryProcessor for every request, to resolve threading issues currently being experienced.

### DIFF
--- a/common-rest/src/test/java/org/uniprot/api/rest/service/query/UniProtQueryProcessorTest.java
+++ b/common-rest/src/test/java/org/uniprot/api/rest/service/query/UniProtQueryProcessorTest.java
@@ -1,12 +1,15 @@
 package org.uniprot.api.rest.service.query;
 
+import nl.altindag.log.LogCaptor;
 import org.apache.lucene.queryparser.flexible.core.QueryNodeException;
 import org.apache.lucene.queryparser.flexible.core.QueryNodeParseException;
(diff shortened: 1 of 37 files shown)


common-rest/src/test/java/org/uniprot/api/rest/service/query/UniProtQueryProcessorTest.java
```diff
@@ -33,19 +37,53 @@
...
+
+        executorService.shutdown();
+        try {
+            if (!executorService.awaitTermination(800, TimeUnit.MILLISECONDS)) {
+                executorService.shutdownNow();
+            }
+        } catch (InterruptedException e) {
+            executorService.shutdownNow();
+        }
     }
```
ahmadshadab, line 88: What does it test? Should we assert some expected result in the test case?

ahmadshadab (approved):

common-rest/src/test/java/org/uniprot/api/rest/service/query/UniProtQueryProcessorTest.java
eddturner: totally agree! i forgot about this -- and just added a proper test

ahmadshadab: cool. please merge.

PR #214, opened 2021-11-15
fix Not Operator

LeonardoGonzales:


uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
```diff
@@ -227,12 +230,15 @@ private boolean needsToFilterIsoform(String query, boolean isIncludeIsoform) {
...
     }
 
-    private boolean needToFilterActiveEntries(UniProtKBSearchRequest uniProtRequest) {
+    private boolean isSearchAll(UniProtKBSearchRequest uniProtRequest) {
         return "*".equals(uniProtRequest.getQuery().trim())
                 || "(*)".equals(uniProtRequest.getQuery().trim())
                 || "*:*".equals(uniProtRequest.getQuery().trim())
-                || "(*:*)".equals(uniProtRequest.getQuery().trim())
-                || SolrQueryUtil.hasNegativeTerm(uniProtRequest.getQuery());
+                || "(*:*)".equals(uniProtRequest.getQuery().trim());
```
eddturner, line 237: use strip() (in all of the `trim()`s), since strip removes all whitespace characters in unicode https://stacktraceguru.com/java/remove_spaces_from_string#trim_vs_strip

eddturner (approved):
1 change, then merge

PR #217, opened 2021-11-19
TRM-26941: add other names as default search field in taxonomy

LeonardoGonzales:
TRM-26941


ahmadshadab (approved):
very simple input string for other name in the test. :-)

PR #219, opened 2021-11-25
update xml schemas and other xml changes

ahmadshadab:
See TRM-27006 for more details. Basically, it will unblock Jerven.
We cannot update the copyright of UniRef unless we talk  o PIR. See another ticket for that 
 task https://www.ebi.ac.uk/panda/jira/browse/TRM-27009


uniparc-rest/src/main/java/org/uniprot/api/uniparc/output/converter/UniParcXmlMessageConverter.java
```diff
@@ -24,7 +26,7 @@
     public UniParcXmlMessageConverter(String version) {
         super(UniParcEntry.class, UNIPARC_XML_CONTEXT);
         converter = new UniParcEntryConverter();
-        header = UNIPARC_XML_HEADER;
+        header = XML_DECLARATION + UNIPARC_XML_SCHEMA;
         if (Utils.notNullNotEmpty(version)) {
             String versionAttrib = " version=\"" + version + "\"" + ">\n";
             header = header.replace(">\n", versionAttrib);
```
LeonardoGonzales, line 32: We have 2 closing tags for the header now.
I have a feeling that this code might replace twice, and this is not what we want, right?

uniref-rest/src/main/java/org/uniprot/api/uniref/output/converter/UniRefXmlMessageConverter.java
```diff
@@ -13,18 +14,12 @@
...
     private String header;
 
     public UniRefXmlMessageConverter(String version, String releaseDate) {
-        super(UniRefEntry.class, XML_CONTEXT);
+        super(UniRefEntry.class, ConverterConstants.UNIREF_XML_CONTEXT);
         converter = new UniRefEntryConverter();
-        header = HEADER_PREFIX;
+        header = ConverterConstants.XML_DECLARATION + ConverterConstants.UNIREF_XML_SCHEMA;
         if ((version != null) && (!version.isEmpty())) {
             header += " version=\"" + version + "\"";
```
LeonardoGonzales, line 24: Check the replace here as well

eddturner (approved):

PR #220, opened 2021-11-30
Xml validation

eddturner:


LeonardoGonzales (approved):
Excellent. 
I just have a bit of concern accessing external files. It might cause the build to break if the external resource is not available. 
Do you think we could use the xsd file that we currently have in xml-core resources folder? This proposed approach could cause a sync issue with the actual xsd, so, not a perfect world in both cases. :-(

PR #222, opened 2021-12-06
TRM-27044: Improve publication category facet

LeonardoGonzales:
Fixes applied for (https://www.ebi.ac.uk/panda/jira/browse/TRM-27044)

Renamed PublicationCategory.java enum label for UniProtKB related publications.
Allow to display all values for category facet, not only top 5 (changed publication.facet.properties)
Add Unclassified value when we do not have any category associated with the publication 
For community publications we were not mapping Subcellular location category. I changed the logic to allow case insensitive input. (Community input is "Subcellular location" and Computationally mapped input is "Subcellular Location")


uniprotkb-rest/src/main/resources/publication.facet.properties
```diff
@@ -11,6 +11,7 @@ facet.publication.types.value.3=UniProtKB unreviewed (TrEMBL)
 
 facet.publication.categories.label=Category
 facet.publication.categories.allowmultipleselection=true
+facet.publication.categories.limit=15
```
ahmadshadab, line 15: Leo, correct me if i am wrong. I can see there are 10 enums in PublicationCategory. If we include Unclassified then the count will become 11. Should we not set `facet.publication.categories.limit=11`.
Could you please add a test case to verify if we are getting more than 5 facets?

PR #229, opened 2022-01-10
Search optimisations

eddturner:


common-rest/src/main/resources/uniprotkb-query.config
```diff
@@ -3,6 +3,7 @@ ec:{query}^1.1
 organism_name:{query}^2.0
 protgene_default:{query}^3.0
 id_default:{query}^10.0
+id_inactive:{query}^10.0
```
LeonardoGonzales, line 6: Hi Edd, I might have missed something, but I did not understand why we are boosting id_inactive. Can you please help me to understand?

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/UniProtKBSearchControllerIT.java
```diff
@@ -504,354 +507,478 @@ void canNotReturnFacetInformationForXML() throws Exception {
...
-                        .andExpect(status().is(HttpStatus.OK.value()))
-                        .andExpect(
-                                header().string(HttpHeaders.CONTENT_TYPE, APPLICATION_JSON_VALUE))
-                        .andExpect(jsonPath("$.results.size()", is(1)))
-                        .andExpect(jsonPath("$.results.*.primaryAccession", contains("Q00007")));
-            }
+    @ParameterizedTest
+    @Tag("TRM_27083")
+    @ValueSource(strings = {"reviewed", "unreviewed"})
+    void inactiveEntry_canSearchForFullIDs(String type) throws Exception {
```
LeonardoGonzales, line 545: Edd, Do you think we could use InactiveEntryMocker to add inactive entries? I think  (I might be wrong) there is no inactive entry in this test.
I think the same thing is happening with other tests.

common-rest/src/main/resources/uniprotkb-query.config
eddturner: Good point -- I just treated it as a normal ID in search, but perhaps we shouldn't ... I'm not sure what the value should be. Maybe I'll just leave it unboosted?

The reason the tests work is that it is now a separate field, but you are right -- it is less important and should be boosted less. 

Any suggestions? 2? Or not boosted?

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/UniProtKBSearchControllerIT.java
eddturner: Gosh -- I named the test wrong. You are right -- it's not related to inactive entries. But is for https://www.ebi.ac.uk/panda/jira/browse/TRM-27083 and all the conditions required by Elisabeth/Maria, etc. I'll rename the test.

common-rest/src/main/resources/uniprotkb-query.config
LeonardoGonzales: I think, not boosted, right? because inactive entries are not important. if we find any, it should be the last to appear, does it make sense to you?

PR #235, opened 2022-01-19
Changing existing content types to ones displayable in a browser: https://www.ebi.ac.uk/panda/jira/browse/TRM-27294

eddturner:


common-rest/src/main/java/org/uniprot/api/rest/output/UniProtMediaType.java
```diff
@@ -49,15 +48,25 @@
...
                     MediaType.APPLICATION_XML);
 
+    private static final String FORMAT_PARAMETER = "format";
     private static HashBiMap<MediaType, String> mediaTypeExtensionMap = HashBiMap.create();
 
     static {
         mediaTypeExtensionMap.put(FF_MEDIA_TYPE, "txt");
         mediaTypeExtensionMap.put(XLS_MEDIA_TYPE, "xlsx");
         mediaTypeExtensionMap.put(RDF_MEDIA_TYPE, "rdf");
+        mediaTypeExtensionMap.put(MARKDOWN_MEDIA_TYPE, "md");
```
LeonardoGonzales, line 58: Great catch!!!!

common-rest/src/test/java/org/uniprot/api/rest/output/UniProtMediaTypeTest.java
```diff
@@ -17,44 +21,24 @@
...
+        XLS_MEDIA_TYPE_VALUE + ",xlsx",
+        FASTA_MEDIA_TYPE_VALUE + ",fasta",
+        GFF_MEDIA_TYPE_VALUE + ",gff",
+        OBO_MEDIA_TYPE_VALUE + ",obo",
+        RDF_MEDIA_TYPE_VALUE + ",rdf",
+        MARKDOWN_MEDIA_TYPE_VALUE + ",md",
+        APPLICATION_XML_VALUE + ",xml",
+        APPLICATION_JSON_VALUE + ",json"
+    })
+    void checkMediaTypeFileExtensionRoundTrip(String mediaTypeStr, String extension) {
```
LeonardoGonzales, line 38: Excellent!!!!

common-rest/src/main/java/org/uniprot/api/rest/output/converter/AbstractEntityIdHttpMessageConverter.java
```diff
@@ -25,7 +25,7 @@
 
     @Override
     public boolean canWrite(@Nullable Type type, Class<?> clazz, @Nullable MediaType mediaType) {
-        return this.canWrite(mediaType) && MessageConverterContext.class.isAssignableFrom(clazz);
+        return validMediaType(mediaType) && MessageConverterContext.class.isAssignableFrom(clazz);
```
LeonardoGonzales, line 28: I think this method is never used. because we are not calling it from super class anymore

common-rest/src/main/java/org/uniprot/api/rest/output/converter/AbstractUUWHttpMessageConverter.java
```diff
@@ -71,14 +70,33 @@ public boolean canRead(
...
+                if (mediaType.equals(supportedMediaType)) {
+                    return true;
+                }
+            }
+        }
+
+        // the following (taken from
+        // org.springframework.http.converter.AbstractHttpMessageConverter)
+        // can be simplified but this way is more readable.
+        if (mediaType == null || MediaType.ALL.equalsTypeAndSubtype(mediaType)) {
```
LeonardoGonzales, line 90: Can you please help me to understand this bit, What is not clear for me is this mediaType == null.

I was thinking that this if would be 
if (mediaType != null && MediaType.ALL.equalsTypeAndSubtype(mediaType))

I was not able to spot a scenario that we write a content when we do not know the media type (mediaType == null)

Sorry, if this is a dummy question.

PR #237, opened 2022-01-20
support obo in keyword search api

ahmadshadab:


ahmadshadab:
See https://www.ebi.ac.uk/panda/jira/browse/TRM-27284 for details

LeonardoGonzales (approved):
Approved, just double checking, we do not want to add obo in /id endpoint, right? just search, right?

ahmadshadab:
> Approved, just double checking, we do not want to add obo in /id endpoint, right? just search, right?

id endpoint already have obo format.

PR #239, opened 2022-02-07
limiting streamed request concurrency

eddturner:
https://www.ebi.ac.uk/panda/jira/browse/TRM-27371


LeonardoGonzales (commented):
I think we also need to apply it for uniref, uniparc and id-mapping, right? Maybe this will be another task. not sure...

uniref-rest/src/main/java/org/uniprot/api/uniref/output/UniRefMessageConverterConfig.java
```diff
@@ -45,7 +45,7 @@
 @Getter
 @Setter
 public class UniRefMessageConverterConfig {
```
LeonardoGonzales, line 47: For UniRef we have to create Gatekeeper bean as well, right?

idmapping-rest/src/main/java/org/uniprot/api/idmapping/service/config/IdMappingConfig.java
```diff
@@ -23,7 +23,7 @@
 @ConfigurationProperties(prefix = "id.mapping.job")
 public class IdMappingConfig {
```
LeonardoGonzales, line 24: For idmapping we have to create Gatekeeper bean as well, right?

idmapping-rest/src/main/java/org/uniprot/api/idmapping/output/MessageConverterConfig.java
```diff
@@ -59,7 +59,7 @@
 @Getter
 @Setter
 public class MessageConverterConfig {
-    private TaskExecutorProperties taskExecutor = new TaskExecutorProperties();
+    private StreamConcurrencyProperties taskExecutor = new StreamConcurrencyProperties();
```
LeonardoGonzales, line 62: For idmapping we have to create Gatekeeper bean as well, right?

LeonardoGonzales (commented):
Few more comments added

support-data-rest/pom.xml
```diff
@@ -25,7 +25,7 @@
 		<dependency>
 			<groupId>${project.groupId}</groupId>
 			<artifactId>uniprot-config</artifactId>
-			<version>${project.parent.version}</version>
+			<version>${uniprot-store.version}</version>
```
LeonardoGonzales, line 28: I think it is better keep parent version, right?

PR #245, opened 2022-03-10
Load *query.config files from a directory, or failing that, the classpath

eddturner:


common-rest/src/main/java/org/uniprot/api/common/repository/search/SolrQueryConfigFileReader.java
```diff
@@ -18,8 +21,28 @@
 @Slf4j
 public class SolrQueryConfigFileReader {
     private static final String COMMENT_PREFIX = "#";
+    private static final String SPRING_CONFIG_LOCATION = "spring.config.location";
     private final SolrQueryConfig.SolrQueryConfigBuilder builder;
```
ahmadshadab, line 25: Where and how are we setting spring.config.location?

ahmadshadab (commented):
Just one small clarification see my other comment

common-rest/src/main/java/org/uniprot/api/common/repository/search/SolrQueryConfigFileReader.java
eddturner: in the Dockerfile, i.e., the container uses `spring.config.location` to read app properties from config/meta/

ahmadshadab: yes, all good. please merge

PR #247, opened 2022-03-10
update to use outgoing.ebi.ac.uk email server

LeonardoGonzales:


eddturner (approved):
awesome, you found a way!

PR #250, opened 2022-03-31
improve help search to find "How can I"

LeonardoGonzales:


help-centre-rest/src/main/java/org/uniprot/api/help/centre/service/HelpCentreService.java
```diff
@@ -45,6 +62,28 @@ public HelpCentreService(
...
+    public QueryResult<HelpCentreEntry> search(SearchRequest request) {
+        HelpCentreSearchRequest searchRequest = (HelpCentreSearchRequest) request;
+        String query = searchRequest.getQuery();
+        if(simpleSearch(query)) {
+            //default simple search (query OR "query") will boost exact search
+            searchRequest.setQuery(query + " OR \"" + query + "\"");
+        }
+        QueryResult<HelpCentreEntry> result = super.search(searchRequest);
+        if(simpleSearch(query) && Utils.notNullNotEmpty(result.getSuggestions())){
+            searchRequest.setQuery(query);
```
ahmadshadab, line 75: Why do we need to call solr if there is suggestion? Can we not use the first response? 
Please demonstrate this scenario in a test.

PR #254, opened 2022-04-12
Rdf service integration

ahmadshadab:


support-data-rest/src/main/resources/application.properties
```diff
@@ -44,37 +44,37 @@ spring.data.solr.socketTimeout=3600000
 solr.query.batchSize=100
 
 ############################### Cross Ref(RDF Server) Streaming properties ###############################
-xref.streamer.rdf.requestUrl=https://www.uniprot.org/database/?
+xref.streamer.rdf.requestUrl=http://10.49.3.120:30836/databases/rdf/{dbIds}
```
eddturner, line 47: needed? I presume these addresses will change. Also, we could use the service names used in k8s in the app properties used in the k8s service, no?

ahmadshadab: Hi Edd,
Thanks for looking into it. But you are right, we need to remove k8s specific values.

PR #259, opened 2022-04-25
Result reordering for advanced search

eddturner:


common-rest/src/test/java/org/uniprot/api/common/repository/search/SolrRequestConverterTest.java
```diff
@@ -34,6 +40,51 @@
...
+        UniProtQueryProcessor processor =
+                UniProtQueryProcessor.newInstance(
+                        new BoostQueryNodeProcessorPipeline(defaultTerms));
+        String query = "a b (c OR f:g)";
+        System.out.println("Query is: " + query);
+        processor.processQuery(query);
+
+        System.out.println("Default Terms after processing:");
+        defaultTerms.stream().forEach(System.out::println);
+    }
```
ahmadshadab, line 58: I wonder why are you printing on console. Should we not verify with assert/

ahmadshadab (approved):
Wow, this is amazing change. Please remmeber to make the k8s config changes.

PR #260, opened 2022-04-28
use spring internal jhcache cachemanager with given ehcache.xml

ahmadshadab:


eddturner (approved):
looks good, thanks for this

PR #266, opened 2022-06-07
Allow ID (TRM-27995) and accession with version (TRM-27996)

LeonardoGonzales:
Allow ID (TRM-27995) and accession with version (TRM-27996) for uniprotkb accession endpoint

### DIFF
--- a/uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/UniProtKBController.java
+++ b/uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/UniProtKBController.java
@@ -4,6 +4,7 @@ import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 import static org.springframework.http.MediaType.APPLICATION_XML_VALUE;
 import static org.uniprot.api.rest.output.UniProtMediaType.*;
 import static org.uniprot.api.rest.output.context.MessageConverterContextFactory.Resource.UNIPROTKB;
+import static org.uniprot.api.rest.output.header.HeaderFactory.createHttpSearchHeader;
 import static org.uniprot.api.uniprotkb.controller.UniProtKBController.UNIPROTKB_RESOURCE;
 
 import java.util.Optional;
@@ -16,15 +17,19 @@ import javax.validation.constraints.Pattern;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.context.request.async.DeferredResult;
+import org.springframework.web.servlet.support.ServletUriComponentsBuilder;
 import org.uniprot.api.common.concurrency.Gatekeeper;
 import org.uniprot.api.common.repository.search.QueryResult;
 import org.uniprot.api.rest.controller.BasicSearchController;
+import org.uniprot.api.rest.output.UniProtMediaType;
 import org.uniprot.api.rest.output.context.MessageConverterContext;
 import org.uniprot.api.rest.output.context.MessageConverterContextFactory;
 import org.uniprot.api.rest.request.IdsSearchRequest;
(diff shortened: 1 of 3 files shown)


uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/UniProtKBController.java
```diff
@@ -412,4 +424,37 @@ private void setPreviewInfo(UniProtKBSearchRequest searchRequest, boolean previe
...
+        String uniProtPath = ServletUriComponentsBuilder.fromCurrentRequest().build().getPath();
+        String uniSavePath = "";
+        if (uniProtPath != null) {
+            String format = UniProtMediaType.getFileExtension(contentType);
+            String accession = accessionOrId.substring(0, accessionOrId.indexOf('.'));
+            String version = accessionOrId.substring(accessionOrId.indexOf('.') + 1);
+            uniSavePath = uniProtPath.replace("uniprotkb", "unisave");
+            uniSavePath = uniSavePath.substring(0, uniSavePath.lastIndexOf('/') + 1) + accession;
+            uniSavePath = uniSavePath +"?from=" + accessionOrId + "&versions=" + version + "&format=" + format;
+        }
```
ahmadshadab, line 440: it would be bit optimal if we use StringBuilder instead of immutable String for uniSavePath

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
```diff
@@ -135,6 +139,36 @@ public UniProtKBEntry findByUniqueId(String accession, String fields) {
...
 
+    public String findAccessionByProteinId(String proteinId) {
+        try {
+            SolrRequest solrRequest =
+                    SolrRequest.builder()
+                            .query(PROTEIN_ID + ":" + proteinId.toUpperCase() + " AND  " + IS_ISOFORM + ":false")
+                            .sorts(solrSortClause.getSort(null))
+                            .rows(NumberUtils.INTEGER_TWO)
+                            .build();
+            QueryResult<UniProtDocument> queryResult = repository.searchPage(solrRequest, null);
```
ahmadshadab, line 150: I believe, you are getting 2 rows as optimisation which is good. But in the below code, you are filtering non-active entry. Is it possible that the first two result has 1 active and 1 inactive but the 3rd (or 3rd onwards) entry might also exist for the same protein id which is active too. Is this case possible?

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/UniProtKBByAccessionControllerIT.java
```diff
@@ -397,6 +397,147 @@ void searchForDeletedInactiveEntriesReturnItself() throws Exception {
...
+
+        // then
+        response.andDo(log())
+                .andExpect(status().is(HttpStatus.SEE_OTHER.value()))
+                .andExpect(header().string(HttpHeaders.CONTENT_TYPE, APPLICATION_JSON_VALUE))
+                .andExpect(content().contentTypeCompatibleWith(APPLICATION_JSON_VALUE))
+                .andExpect(header().string(HttpHeaders.LOCATION, "/uniprotkb/I8FBX2?from=I8FBX2_YERPE"));
+    }
+
     @Override
```
ahmadshadab, line 541: It would be good if we could add more negative test cases.
1. P12345..txt (regex should catch it)
2. redirect for other formats (e.g. xml)
3. when one protein id is mapped to more than one entries
4. some non-existing version

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
LeonardoGonzales: As far as I understood, we can only have one active ID.
There is the Merged or Demerged use case that we will have the same ID as inactive as well. (this is the 2 entry use case). This is why we filter only active one. For us, the active one is important. Does it make sense?

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/UniProtKBByAccessionControllerIT.java
LeonardoGonzales: 1. Good point, I will create a UnitTest for the regex
2. This is other code, right? (probably covered by other test) In the controller we receive the correct MediaType and use it. Does it make sense?
3. It is covered in the Merged and Demerged integration test
4. I think this is other code as well, right? (unisave code, right?)

LeonardoGonzales: I just pushed the unit tests for FieldRegexConstants (FieldRegexConstantsTest)

PR #281, opened 2022-07-15
merge lit_organisation field into lit_author

LeonardoGonzales:


ahmadshadab (approved):
This is an optional comment.
Are we not verifying the author and organisation in rest api response in any of the integ test?
It would be better if you could add couple of assert statement.

PR #283, opened 2022-07-22
TRM-28235: enable search and get by id for canonical isoform

LeonardoGonzales:
TRM-28235: enable search and get by id for canonical isoform that does not have isoforms

### DIFF
--- a/uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
+++ b/uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
@@ -49,6 +49,7 @@ public class UniProtEntryService
     public static final String ACCESSION = "accession_id";
     public static final String PROTEIN_ID = "id";
     public static final String IS_ISOFORM = "is_isoform";
+    public static final String CANONICAL_ISOFORM = "-1";
     private final UniProtEntryQueryResultsConverter resultsConverter;
     private final SolrQueryConfig solrQueryConfig;
     private final UniProtQueryProcessorConfig uniProtQueryProcessorConfig;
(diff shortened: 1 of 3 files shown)


uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
```diff
@@ -119,12 +120,14 @@ protected UniProtQueryProcessorConfig getQueryProcessorConfig() {
...
             List<ReturnField> fieldList = OutputFieldsParser.parse(fields, returnFieldConfig);
-            SolrRequest solrRequest =
-                    SolrRequest.builder()
-                            .query(ACCESSION + ":" + accession.toUpperCase())
-                            .rows(NumberUtils.INTEGER_ONE)
-                            .build();
+            accession = accession.toUpperCase();
+            SolrRequest solrRequest = buildSolrRequest(accession);
             Optional<UniProtDocument> optionalDoc = repository.getEntry(solrRequest);
+            if (accession.endsWith(CANONICAL_ISOFORM) && optionalDoc.isEmpty()) {
```
rizwan-ishtiaq, line 126: Correct me if i am wrong.
accession will never be null
canonical isoform is always ends with xxx-1 means xxx-2 or xxx-3 never be canonical isoform
we have index solr with canonical isoform

rizwan-ishtiaq (approved):
looks good to me

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
LeonardoGonzales: accession will never be null: correct, accession is a required parameter
Yes, basically the uniprot entry P12345 and P12345-1 are the same thing, the user can request both ways, and we need to return P12345. (first request will fail, and we remove the -1 and query again)
When we have entries that has canonical + isoforms (P21802, P21802-1, P21802-2, ...) (we are indexing it as 2 different entries in solr and voldemort), in this case, the first request will return it.
Does it make sense?

PR #287, opened 2022-08-15
TRM-28329: improve citation id Regex

LeonardoGonzales:
TRM-28329: improve citation id Regex


jieluo321 (changes requested):
AbstractCitationImpl.generateHash can return any length of string
for example crc64="0000000000004051";
String base32Hash = new BigInteger(crc64, 16).toString(32).toUpperCase();
return G2H

PR #290, opened 2022-08-25
TRM-28253 And TRM-28365: Add id-mapping subsequences of proteins and …

LeonardoGonzales:
…allow accessions with version

### DIFF
--- a/common-rest/src/main/java/org/uniprot/api/rest/output/context/MessageConverterContext.java
+++ b/common-rest/src/main/java/org/uniprot/api/rest/output/context/MessageConverterContext.java
@@ -35,6 +35,7 @@ public class MessageConverterContext<T> {
     private Collection<String> failedIds;
     private Collection<ProblemPair> warnings;
     private boolean isLargeDownload;
+    private boolean subsequence;
 
     MessageConverterContext<T> asCopy() {
         return MessageConverterContext.<T>builder()
@@ -52,6 +53,7 @@ public class MessageConverterContext<T> {
                 .suggestions(this.suggestions)
                 .warnings(this.warnings)
                 .isLargeDownload(this.isLargeDownload)
+                .subsequence(subsequence)
                 .build();
     }
 }
--- a/idmapping-rest/src/main/java/org/uniprot/api/idmapping/controller/request/uniprotkb/UniProtKBIdMappingBasicRequest.java
+++ b/idmapping-rest/src/main/java/org/uniprot/api/idmapping/controller/request/uniprotkb/UniProtKBIdMappingBasicRequest.java
@@ -6,13 +6,11 @@ import lombok.Data;
 import lombok.EqualsAndHashCode;
 
 import org.uniprot.api.idmapping.controller.request.IdMappingPageRequest;
+import org.uniprot.api.rest.output.UniProtMediaType;
 import org.uniprot.api.rest.request.QueryFieldMetaReaderImpl;
 import org.uniprot.api.rest.request.ReturnFieldMetaReaderImpl;
 import org.uniprot.api.rest.request.SortFieldMetaReaderImpl;
-import org.uniprot.api.rest.validation.ValidReturnFields;
-import org.uniprot.api.rest.validation.ValidSolrQueryFields;
-import org.uniprot.api.rest.validation.ValidSolrQuerySyntax;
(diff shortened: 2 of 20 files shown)


idmapping-rest/src/main/java/org/uniprot/api/idmapping/output/converter/uniprotkb/UniProtKBEntryPairFastaMessageConverter.java
```diff
@@ -1,13 +1,22 @@
...
+import org.uniprot.api.rest.output.context.MessageConverterContext;
+import org.uniprot.core.fasta.UniProtKBFasta;
+import org.uniprot.core.fasta.impl.UniProtKBFastaBuilder;
 import org.uniprot.core.parser.fasta.uniprot.UniProtKBFastaParser;
 import org.uniprot.core.uniprotkb.UniProtKBEntry;
 
 public class UniProtKBEntryPairFastaMessageConverter
         extends AbstractEntryPairFastaConverter<UniProtKBEntryPair, UniProtKBEntry> {
+
+    private static final ThreadLocal<Boolean> TL_SUBSEQUENCE = new ThreadLocal<>();
```
rizwan-ishtiaq, line 18: We don't need to use ThreadLocal here
i have seen usage, we are creating new instance and instance it self is not shared by multi thread

common-rest/src/main/java/org/uniprot/api/rest/output/context/MessageConverterContext.java
```diff
@@ -52,6 +53,7 @@ MessageConverterContext<T> asCopy() {
                 .suggestions(this.suggestions)
                 .warnings(this.warnings)
                 .isLargeDownload(this.isLargeDownload)
+                .subsequence(subsequence)
```
rizwan-ishtiaq, line 56: Write test - 
1. by default subseq will be false
2. you can set it false as well if true

idmapping-rest/src/main/java/org/uniprot/api/idmapping/controller/request/uniprotkb/UniProtKBIdMappingBasicRequest.java
```diff
@@ -51,6 +49,20 @@ public class UniProtKBIdMappingBasicRequest extends IdMappingPageRequest {
...
 
+    @Parameter(description = "Flag to write subsequences. Only accepted in fasta format")
+    @Pattern(
+            regexp = "true|false",
+            flags = {Pattern.Flag.CASE_INSENSITIVE},
+            message = "{search.invalid.subsequence}")
+    @ValidContentTypes(
+            contentTypes = {UniProtMediaType.FASTA_MEDIA_TYPE_VALUE},
+            message = "{search.invalid.contentType.subsequence}")
+    private String subSequence;
```
rizwan-ishtiaq, line 60: I might miss it - but did you test
asking for sub sequence for any other format should not allowed

idmapping-rest/src/main/java/org/uniprot/api/idmapping/output/converter/uniprotkb/UniProtKBEntryPairFastaMessageConverter.java
```diff
@@ -17,7 +26,45 @@ public UniProtKBEntryPairFastaMessageConverter(Gatekeeper downloadGatekeeper) {
...
+                .id(entry.getPrimaryAccession().getValue())
+                .entryType(entry.getEntryType())
+                .uniProtkbId(range)
+                .sequence(getSubSequence(range, entry.getSequence().getValue()))
+                .build();
+    }
+
+    private String getSubSequence(String range, String sequence) {
+        String[] rangeValues = range.split("-");
+        int sequenceStart = Integer.parseInt(rangeValues[0]);
```
rizwan-ishtiaq, line 66: someone could send invalid request 
passing char instad of number
e-g instead of number 0 passing alphabet o

rizwan-ishtiaq (approved):
It is difficult to focus on test cases and possible senerios  if single PR contain multiple busniess logic changes. 

Although you have put lot of test, specially to test the logic. But it would be better to write test for those areas where you are parsing/validating input

PR #298, opened 2022-09-26
filter separator only if it is accession

ahmadshadab:


LeonardoGonzales (approved):

idmapping-rest/src/main/java/org/uniprot/api/idmapping/service/job/PIRJobTask.java
```diff
@@ -27,6 +30,11 @@ protected IdMappingResult processTask(IdMappingJob job) {
             return IdMappingResult.builder()
                     .error(new ProblemPair(REST_EXCEPTION_CODE, restException.getMessage()))
                     .build();
+        } catch (Exception ex) {
+            log.error("Error while processing PIR response {}", ex.getCause());
```
LeonardoGonzales, line 34: I think it would be easier to trace this issue if we add the jobId in the log error message

ahmadshadab: Thanks.. I've added job id in the log message.

PR #300, opened 2022-10-10
TRM-28565:add missing regex validation to boolean fields

LeonardoGonzales:


ahmadshadab (approved):
Do we have any similar test case in publication where we need to add value for is_large_scale/>

LeonardoGonzales:
This is a facet search and we have facetValidation annotation, right?

PR #303, opened 2022-10-26
TRM-28516: search WP_168893289 without double quote does not work

LeonardoGonzales:


ahmadshadab (approved):
I am approving this but is it the perfect solution? We are changing the user's query.

LeonardoGonzales:
This is how the current website does it. (We might want to try to improve it later)

PR #307, opened 2022-11-02
Leading wildcard warning

ahmadshadab:


uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
```diff
@@ -314,4 +319,14 @@ private boolean needToAddActiveFilter(UniProtKBSearchRequest uniProtRequest) {
     private String getQueryFieldName(String active) {
         return searchFieldConfig.getSearchFieldItemByName(active).getFieldName();
     }
+
+    private ProblemPair getLeadingWildcardIgnoredWarning(String query) {
+        if (SolrQueryUtil.ignoreLeadingWildcard(
+                query, this.uniProtQueryProcessorConfig.getLeadingWildcardFields())) {
+            return new ProblemPair(
+                    41,
+                    "Leading wildcard (*, ?) was removed for this search. Please check the help page for more information on using wildcards on queries.");
```
LeonardoGonzales, line 328: I think we have an enum to group all these messages, right? Do you think we could use it here? An even better solution would be to use message.properties for it.

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
```diff
@@ -102,7 +105,9 @@ public QueryResult<UniProtKBEntry> search(SearchRequest request) {
         QueryResult<UniProtDocument> results =
                 repository.searchPage(solrRequest, request.getCursor());
         List<ReturnField> fields = OutputFieldsParser.parse(request.getFields(), returnFieldConfig);
-        return resultsConverter.convertQueryResult(results, fields);
+        ProblemPair warning = getLeadingWildcardIgnoredWarning(request.getQuery());
```
LeonardoGonzales, line 108: I think this logic needs to be in the parent/generic class, because the leading wildcard logic is used for everything, not only UniProt, we need to make sure we add the warning for everything, Does it make sense to you?

LeonardoGonzales (commented):
Let me know if you have any doubt

LeonardoGonzales (approved):

PR #308, opened 2022-11-08
TRM-28741: Apply field optimisation only for single term search

LeonardoGonzales:


common-rest/src/test/java/org/uniprot/api/rest/service/query/UniProtQueryProcessorTest.java
```diff
@@ -209,21 +209,16 @@ void useAndAsDefaultInNestedQuery() {
...
-        String ACC = "P12345";
-        String pre = "a OR ( b AND ( +c:something AND -d:something ) AND ( ";
-        String post = " OR range:[1 TO 2] OR range:[1 TO *] OR range:[* TO 1] ) )";
-        String query = pre + ACC + post;
-        String processedQuery = processor.processQuery(query);
-        assertThat(processedQuery, is(pre + FIELD_NAME + ":" + ACC + post));
+    void doNotOptimisesPartOfQuery() {
+        String processedQuery = processor.processQuery("a OR P12345");
+        assertThat(processedQuery, is("a OR P12345"));
     }
```
ahmadshadab, line 221: Could you please add a test case with accession in the start of the query string e.g. P12345 AND cancer

ahmadshadab (approved):
Please merge after adding the test. Sorry for the late review. I totally missed it.

PR #311, opened 2022-12-07
Rdf service integration

ahmadshadab:


uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/service/UniProtEntryService.java
```diff
@@ -211,6 +211,13 @@ public String findAccessionByProteinId(String proteinId) {
     public Stream<String> streamRDF(UniProtKBStreamRequest streamRequest) {
         SolrRequest solrRequest =
                 createSolrRequestBuilder(streamRequest, solrSortClause, solrQueryConfig).build();
+        if (UniProtKBRequestUtil.needsToFilterIsoform(
```
LeonardoGonzales, line 214: Is there any test case for the isoform usecase?

uniprotkb-rest/src/main/resources/application.properties
```diff
@@ -31,23 +31,26 @@ spring.resources.add-mappings=true
...
 #spring.data.solr.zkHost=localhost:2181
 #spring.data.solr.httphost=http://wp-np2-b4:8983/solr/
+#spring.data.solr.kb.httphost=http://10.3.44.12:31607/solr/
 #spring.data.solr.httphost=https://www.ebi.ac.uk/uniprot/services/search/
 #spring.data.solr.httphost=http://localhost:8082/solr/
 
 spring.data.solr.kb.username=solr_admin
 spring.data.solr.kb.password=nimda
-spring.data.solr.kb.connectionTimeout=20000
+spring.data.solr.kb.connectionTimeout=120000
```
LeonardoGonzales, line 49: Maybe we should keep 20000, because the new way is very fast, right?

uniprotkb-rest/src/main/resources/application.properties
```diff
@@ -76,19 +80,19 @@ voldemort.uniprot.storeName=uniprotkb
 streamer.uniprot.storeBatchSize=5000
 streamer.uniprot.storeFetchMaxRetries=5
 streamer.uniprot.storeFetchRetryDelayMillis=500
-#streamer.uniprot.zkHost=wp-np2-b9.ebi.ac.uk:2191,ves-hx-ee.ebi.ac.uk:2191,ves-hx-c9.ebi.ac.uk:2191
-#streamer.uniprot.zkHost=wp-np2-b3:5191,wp-np2-b4:5191,wp-np2-b5:5191
-streamer.uniprot.zkHost=wp-np3-db:5191
```
LeonardoGonzales, line 81: I think we could keep this line, this way we can easily switch back to the new solr env

uniprotkb-rest/src/main/resources/application.properties
```diff
@@ -110,4 +114,5 @@ ids.max.download.length=10000
 #################### cache control max age in seconds ##################################
 cache.control.max.age=86400
 
-server.tomcat.max-connections=1000
\ No newline at end of file
+server.tomcat.max-connections=1000
+logging.config=
```
LeonardoGonzales, line 118: Do we need this logging.config?

LeonardoGonzales (approved):

PR #312, opened 2022-12-09
TRM-28856: Query field validation with forward slash query always ret…

LeonardoGonzales:
…urn true

### DIFF
--- a/common-rest/src/test/java/org/uniprot/api/rest/validation/QueryFieldValidatorTest.java
+++ b/common-rest/src/test/java/org/uniprot/api/rest/validation/QueryFieldValidatorTest.java
@@ -232,6 +232,28 @@ class QueryFieldValidatorTest {
         assertEquals("taxonomy_id", validator.getErrorFields(ErrorType.VALUE).get(0));
     }
 
+    @Test
+    void isValidWithForwardSlashReturnTrue() {
+        ValidSolrQueryFields validSolrQueryFields = getMockedValidSolrQueryFields();
+        FakeQueryFieldValidator validator = new FakeQueryFieldValidator();
+        validator.initialize(validSolrQueryFields);
+
+        boolean result = validator.isValid("a/b//c", null);
+        assertTrue(result);
+    }
+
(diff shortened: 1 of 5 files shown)


common-rest/src/test/java/org/uniprot/api/rest/validation/QueryFieldValidatorTest.java
```diff
@@ -232,6 +232,28 @@ void isValidInvalidIntegerFieldReturnFalse() {
...
+        FakeQueryFieldValidator validator = new FakeQueryFieldValidator();
+        validator.initialize(validSolrQueryFields);
+
+        boolean result = validator.isValid("invalid:a//b/c", null);
+        assertFalse(result);
+        assertEquals(1, validator.getErrorFields(ErrorType.FIELD).size());
+        assertEquals("invalid", validator.getErrorFields(ErrorType.FIELD).get(0));
+    }
+
     private ValidSolrQueryFields getMockedValidSolrQueryFields() {
```
ahmadshadab, line 257: Leo, do we have a test where field name has a forward slash? If not, i feel, it would be good to add such. What do you think/

ahmadshadab: Please merge after that.. No need to review again.

LeonardoGonzales: I added tests with forward slash values, but the field name with forward slack will fail in the syntax. Not sure if it is necessary

ahmadshadab: oh i thought the issue we noticed in solr prod had forward slash in the field name. Was it not the case? If so please merge

LeonardoGonzales: the forward slash was on the values. I will merge it.

PR #314, opened 2022-12-15
TRM-28888: Allow query optimiser to accept brackets

LeonardoGonzales:


common-rest/src/main/java/org/uniprot/api/rest/service/query/UniProtQueryProcessor.java
```diff
@@ -79,16 +79,17 @@ public String processQuery(String query) {
     }
 
     private String proccessOptimisableFields(String text) {
+        String cleanQuery = text.strip().replaceAll("^\\(|\\)$", "");
         Optional<SearchFieldItem> optionalSearchField =
```
ahmadshadab, line 83: Could you please move the regex to Pattern.compile(regex)? It will be faster

common-rest/src/test/java/org/uniprot/api/rest/service/query/UniProtQueryProcessorTest.java
```diff
@@ -448,6 +448,18 @@ void optimiseDefaultSearchValueLowercase() {
...
+
+    @Test
+    void doNotOptimiseDefaultSearchWithComplexQuery() {
+        String processedQuery = processor.processQuery("(p12345 cancer)");
+        assertThat(processedQuery, is("( p12345 AND cancer )"));
+    }
+
     @Test
     void changeLowercaseAccessionToUppercase() {
         String processedQuery = processor.processQuery("GO:1234567 OR accession:p12345");
```
ahmadshadab, line 465: What if we have query like 
1. (p12345) AND (organism_id:9606) 
2. ((p12345))

ahmadshadab (approved):

PR #316, opened 2023-01-06
TRM-28929: Remove ccev and ftev search fields from advanced search

LeonardoGonzales:


common-rest/src/test/java/org/uniprot/api/rest/validation/QueryFieldValidatorTest.java
```diff
@@ -112,7 +112,7 @@ void isValidBooleanSubQueryReturnTrue() {
                 validator.isValid(
```
ahmadshadab, line 112: there is a reference of a ccev_* in src/test/resources/uniprot/search-fields.json.

PR #324, opened 2023-02-27
Async Download

ahmadshadab:


idmapping-rest/src/main/java/org/uniprot/api/idmapping/controller/request/IdMappingJobRequestToArrayConverter.java
```diff
@@ -0,0 +1,28 @@
...
+import org.uniprot.core.util.Utils;
+
+public class IdMappingJobRequestToArrayConverter implements Function<IdMappingJobRequest, char[]> {
+    @Override
+    public char[] apply(IdMappingJobRequest request) {
+        StringBuilder builder = new StringBuilder();
+        builder.append(request.getFrom().strip().toLowerCase());
+        builder.append(request.getTo().strip().toLowerCase());
+        builder.append(
+                List.of(request.getIds().strip().split(",")).stream()
```
supun-ebi, line 17: Minor Comment: 
You CAN replace List.of(request.getIds().strip().split(",")).stream() with Stream.of(request.getIds().strip().split(","))

(Can be seen from IDE suggestions as well)

idmapping-rest/src/main/java/org/uniprot/api/idmapping/service/impl/IdMappingJobServiceImpl.java
```diff
@@ -74,7 +77,8 @@ public IdMappingJobServiceImpl(
         this.cacheService = cacheService;
         this.pirService = pirService;
         this.jobTaskExecutor = jobTaskExecutor;
-        this.hashGenerator = new HashGenerator();
+        // make it autowired bean TODO
+        this.hashGenerator = new HashGenerator(new IdMappingJobRequestToArrayConverter(), SALT_STR);
```
supun-ebi, line 81: Minor Comment: 
new HashGenerator<>(new IdMappingJobRequestToArrayConverter(), SALT_STR); is more approppriate.
(diamond sign after generic class name)

(Can be seen from IDE suggestions as well)

common-rest/src/main/java/org/uniprot/api/rest/controller/BasicSearchController.java
```diff
@@ -279,6 +286,44 @@ protected MessageConverterContext<T> createStreamContext(
         return context;
     }
 
+    protected ResponseEntity<JobStatusResponse> createAsyncDownloadStatus(
+            DownloadJob job, String url) {
```
supun-ebi, line 290: Here, in this method parameter 'url' has not been used.

Also, I doubt whether this class (BasicSearchController) is the best place to keep this method. I would prefer to have it in a child class s.t. like 'AsyncDownloadableSeachController' or keep it a helper class.

Minor comment: I prefer a name like getAsyncDownloadStatus since it is just wrapping the already created status.

common-rest/src/main/java/org/uniprot/api/rest/download/model/DownloadRequestToArrayConverter.java
```diff
@@ -0,0 +1,25 @@
...
+        builder.append(request.getQuery().strip().toLowerCase());
+        if (request.hasFields()) {
+            builder.append(request.getFields().strip().toLowerCase());
+        }
+        if (request.hasSort()) {
+            builder.append(request.getSort().strip().toLowerCase());
+        }
+
+        if (Objects.nonNull(request.getFormat())) {
+            builder.append(request.getFormat().toString().toLowerCase());
```
supun-ebi, line 21: This might not be your code. Just saw calling string.toString()

common-rest/src/main/java/org/uniprot/api/rest/download/model/HashGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+ * @author sahmad
+ * @created 22/02/2021
+ */
+public class HashGenerator<T> {
+    private static final String ALGORITHM_NAME = "PBKDF2WithHmacSHA1";
+    private static final int ITERATION_COUNT = 16;
+    private static final int KEY_LENGTH = 160;
+    private final byte[] saltBytes;
+
+    private Function<T, char[]> requestToArrayConverter;
```
supun-ebi, line 24: Why this is not final while saltBytes is final?

common-rest/src/main/java/org/uniprot/api/rest/download/queue/QueueConsumerConfig.java
```diff
@@ -0,0 +1,38 @@
...
+ * @created 22/11/2022
+ */
+@Configuration
+@Profile({"asyncDownload"})
+public class QueueConsumerConfig {
+
+    @Bean
+    public MessageListenerContainer messageListenerContainer(
+            ConnectionFactory connectionFactory,
+            @Qualifier("Consumer") MessageListener messageListener,
```
supun-ebi, line 23: I don't like the name 'Consumer'. It doesn't make it specific (If some other comes, that would also be a consumer). 
What if something like UniProtKBMessageListener?

common-rest/src/main/java/org/uniprot/api/rest/download/configuration/RedisConfiguration.java
```diff
@@ -0,0 +1,50 @@
...
+        return new RedissonConnectionFactory(redisson);
+    }
+
+    @Bean(destroyMethod = "shutdown")
+    @Profile("asyncDownload & live")
+    public RedissonClient redisson(@Value("${download.redis.config.file}") Resource configFile)
+            throws IOException {
+        Config config = Config.fromYAML(configFile.getInputStream());
+        RedissonClient client = Redisson.create(config);
+        return client;
```
supun-ebi, line 40: This is an old code I guess, however I think you can return edisson.create(config), without having a redundant variable.

common-rest/src/main/java/org/uniprot/api/rest/download/queue/RabbitMQConfig.java
```diff
@@ -0,0 +1,116 @@
...
+        CachingConnectionFactory connectionFactory =
+                new CachingConnectionFactory(rabbitMQConfigProperties.getHost());
+        connectionFactory.setUsername(rabbitMQConfigProperties.getUser());
+        connectionFactory.setPassword(rabbitMQConfigProperties.getPassword());
+        connectionFactory.setPort(rabbitMQConfigProperties.getPort());
+        return connectionFactory;
+    }
+
+    @Bean
+    public Exchange downloadExchange(RabbitMQConfigProperties rabbitMQConfigProperties) {
```
supun-ebi, line 44: Isn't this download agnostic? In that case, you can stick to a more generic name s.t. directExchange

common-rest/src/main/java/org/uniprot/api/rest/download/queue/RabbitMQConfigProperties.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private int concurrentConsumers;
+
+    private boolean defaultRequeueRejected;
+
+    private int retryDelayInMillis;
+
+    private int retryMaxCount;
+
+    private String retryQueueName;
```
supun-ebi, line 39: I guess it's better to extract out the queue specific properties from general rabbitMQ properties like host, port etc.

common-rest/src/main/java/org/uniprot/api/rest/download/queue/RabbitProducerMessageService.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    private void createDownloadJob(String jobId, DownloadRequest downloadRequest) {
+        DownloadJob.DownloadJobBuilder jobBuilder = DownloadJob.builder();
+        LocalDateTime now = LocalDateTime.now();
+        jobBuilder.id(jobId).status(JobStatus.NEW);
+        String format =
+                Objects.nonNull(downloadRequest.getFormat())
+                        ? downloadRequest.getFormat().toString()
```
supun-ebi, line 81: again string.toString()

common-rest/src/main/java/org/uniprot/api/rest/request/HttpServletRequestContentTypeMutator.java
```diff
@@ -67,8 +68,10 @@ static void orderKeysSoPathVariablesLast(List<String> pathVariables) {
     }
 
     public void mutate(MutableHttpServletRequest request) {
-        mutateAcceptHeader(request);
-        mutateAcceptEncodingHeader(request);
+        if (!request.getRequestURI().endsWith(DOWNLOAD_RUN_SUFFIX)) {
```
supun-ebi, line 71: I would prefer this if check to be performed by the caller and this method only does the mutation part.

common-rest/src/main/java/org/uniprot/api/rest/validation/ValidAsyncDownloadFormats.java
```diff
@@ -0,0 +1,97 @@
...
+    String message() default "{search.invalid.format}";
+
+    Class<?>[] groups() default {};
+
+    Class<? extends Payload>[] payload() default {};
+
+    class FormatsValidator implements ConstraintValidator<ValidAsyncDownloadFormats, String> {
+        private static final Logger LOGGER = LoggerFactory.getLogger(FormatsValidator.class);
+
+        @Autowired private HttpServletRequest request;
```
supun-ebi, line 43: This might not be your code as well, anyway this field does not have any usage.

common-rest/src/main/java/org/uniprot/api/rest/validation/ValidAsyncDownloadFormats.java
```diff
@@ -0,0 +1,97 @@
...
+                    type = UniProtMediaType.getMediaTypeForFileExtension(value).toString();
+                } catch (IllegalArgumentException ile) {
+                    type = "";
+                }
+                if (!getFormats().contains(value) && !getFormats().contains(type)) {
+                    buildUnsupportedFormatErrorMessage(value, contextImpl);
+                    isValid = false;
+                }
+
+                if (!isValid && contextImpl != null) {
```
supun-ebi, line 78: this second condition after '&&' is always true.

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/UniProtKBDownloadController.java
```diff
@@ -0,0 +1,146 @@
...
+                converterContextFactory,
+                downloadTaskExecutor,
+                UNIPROTKB,
+                downloadGatekeeper);
+        this.messageService = messageService;
+        this.jobRepository = jobRepository;
+        this.hashGenerator = new HashGenerator<>(new DownloadRequestToArrayConverter(), SALT_STR);
+    }
+
+    @PostMapping(value = "/run", produces = APPLICATION_JSON_VALUE)
```
supun-ebi, line 75: General comment about the naming: as per the REST naming conventions
(1) there should not be verbs
(2) multi-resource paths are named in the plural

So a suitable name for this is just /downloads or /downloads/new
 And for status something like downloads/{id}/status
 
However I see throughout the application, it is not the case, so I'm fine with it.

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/UniProtKBDownloadController.java
```diff
@@ -0,0 +1,146 @@
...
+    public ResponseEntity<JobSubmitResponse> submitJob(
+            @Valid @ModelAttribute UniProtKBDownloadRequest request,
+            HttpServletRequest httpRequest) {
+        MessageProperties messageHeader = new MessageProperties();
```
supun-ebi, line 79: IMO, I would prefer an intermediate class to do these kinds of processing. It has happened because you directly talk to the repository from the controller itself so that you dont have any other place to write the processing logic.

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/queue/AsyncDownloadIntegTest.java
```diff
@@ -0,0 +1,343 @@
...
+            UniProtKBREST.class,
+            UniProtStoreConfig.class,
+            AsyncDownloadTestConfig.class,
+            RedisConfiguration.class
+        })
+@ExtendWith(SpringExtension.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@WebMvcTest({UniProtKBDownloadController.class})
+@AutoConfigureWebClient
+public class AsyncDownloadIntegTest extends AbstractUniProtKBDownloadIT {
```
supun-ebi, line 73: I am a bit unclear on this name, For me it looks more like RabbitProducerMessageServiceIT

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/queue/AsyncDownloadIntegTest.java
```diff
@@ -0,0 +1,343 @@
...
+    @Autowired private TupleStreamTemplate tupleStreamTemplate;
+    @Autowired private FacetTupleStreamTemplate facetTupleStreamTemplate;
+
+    @Autowired
+    @Qualifier("uniProtKBSolrClient")
+    private SolrClient
+            solrClient; // this is NOT inmemory solr cluster client, see CloudSolrClient in parent
+    // class
+
+    @SpyBean @Autowired
```
supun-ebi, line 83: Why you want to use @Autowired and @SpyBean together?

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/queue/UniProtKBMessageListenerTest.java
```diff
@@ -0,0 +1,153 @@
...
+    void testOnMessageWhenMaxRetryReached() throws IOException {
+        UniProtKBDownloadRequest downloadRequest = new UniProtKBDownloadRequest();
+        downloadRequest.setQuery("field1:value1");
+        downloadRequest.setFormat(MediaType.APPLICATION_JSON.toString());
+        String jobId = UUID.randomUUID().toString();
+        MessageBuilder builder = MessageBuilder.withBody(downloadRequest.toString().getBytes());
+        Message message = builder.setHeader("jobId", jobId).build();
+        DownloadJob downloadJob = DownloadJob.builder().id(jobId).build();
+        this.uniProtKBMessageListener.setMaxRetryCount(0);
+        Assertions.assertDoesNotThrow(() -> this.uniProtKBMessageListener.onMessage(message));
```
LeonardoGonzales, line 117: How do you know that it has returned the maxRetry use case? is there a way to assert it?

common-rest/src/main/java/org/uniprot/api/rest/download/queue/QueueConsumerConfig.java
ahmadshadab: We need to change the Qualifier

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/queue/UniProtKBMessageListenerTest.java
ahmadshadab: We are setting the maxretry to 0(it will be 3(> 0) in prod) and the call will enter into if block of  if (isMaxRetriedReached(message)) {}; in the very first call itself.
There is nothing to assert here except that it doen's throw any exception.

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/AbstractUniProtKBDownloadIT.java
```diff
@@ -0,0 +1,203 @@
...
+        entryBuilder.primaryAccession(acc);
+
+        UniProtKBEntry uniProtKBEntry = entryBuilder.build();
+        UniProtDocument convert = documentConverter.convert(uniProtKBEntry);
+
+        cloudSolrClient.addBean(SolrCollection.uniprot.name(), convert);
+        storeClient.saveEntry(uniProtKBEntry);
+    }
+
+    protected void uncompressFile(Path zippedFile, Path unzippedFile) throws IOException {
```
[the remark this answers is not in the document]
ahmadshadab: moved to testutils

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/queue/AsyncDownloadIntegTest.java
ahmadshadab: thanks @supun-ebi we need only @SpyBean annotation.

common-rest/src/main/java/org/uniprot/api/rest/output/job/DownloadJobDetailResponse.java
```diff
@@ -0,0 +1,32 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/**
+ * @author lgonzales
+ * @since 23/04/2021
+ */
+@Data
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public class DownloadJobDetailResponse implements StreamRequest {
```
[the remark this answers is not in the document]
ahmadshadab: ouch. i don't know it. let me remove the implements

common-rest/src/main/java/org/uniprot/api/rest/download/queue/RabbitMQConfig.java
ahmadshadab: thanks

common-rest/src/main/java/org/uniprot/api/rest/download/configuration/RedisConfiguration.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+import org.springframework.core.io.Resource;
+import org.springframework.data.redis.core.RedisTemplate;
+import org.springframework.data.redis.repository.configuration.EnableRedisRepositories;
+import org.uniprot.api.rest.download.model.DownloadJob;
+
+/**
+ * TODO check possible duplicate code in {@link
```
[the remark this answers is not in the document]
ahmadshadab: I checked. They're different use case. We will have just have one bean with 2 lines of code for idmapping and one for async.

LeonardoGonzales (approved):

PR #330, opened 2023-03-14
TRM-29300: Update to use spark UniProtEntryConverter in Integration t…

LeonardoGonzales:
…ests


common-rest/pom.xml
```diff
@@ -223,13 +223,6 @@
...
 			<scope>test</scope>
 		</dependency>
-		<dependency>
-			<groupId>org.uniprot</groupId>
-			<artifactId>uniprot-indexer-integration-test</artifactId>
-			<version>${uniprot-store.version}</version>
-			<type>test-jar</type>
-			<scope>test</scope>
-		</dependency>
 		<dependency>
```
ahmadshadab, line 226: good God we removed this. thanks

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/TestUtils.java
```diff
@@ -26,51 +23,4 @@ static UniProtKBEntry convertToUniProtEntry(UniProtEntryObjectProxy objectProxy)
     static InputStream getResourceAsStream(String resourcePath) {
         return TestUtils.class.getResourceAsStream(resourcePath);
     }
-
```
ahmadshadab, line 29: Boys scout rule. 👍

ahmadshadab (commented):
I have one comment about usage of old indexer class UniProtEntryConverter. After your changes, we still have two usage of it first in DataStoreManagerTest and other in UniProtSearchEngine. Should we not use the spark one?

LeonardoGonzales:
UniProtSearchEngine is in Store integration test, it is good to do, but I think it is not the priority now. I created a Jira for it: https://www.ebi.ac.uk/panda/jira/browse/TRM-29305

PR #331, opened 2023-03-17
TRM-28917: Accessions endpoint allow canonical isoform for trembl

LeonardoGonzales:
Accessions endpoint not returning data when using entries with only one canonical isoform


common-rest/src/main/java/org/uniprot/api/rest/service/StoreStreamerSearchService.java
```diff
@@ -125,28 +126,41 @@ protected SolrRequest createDownloadSolrRequest(StreamRequest request) {
...
-    private boolean needSolrReturnedAccessions(IdsSearchRequest idsRequest) {
+    private boolean needSolrReturnedAccessions(IdsSearchRequest idsRequest, boolean hasIsoformIds) {
         return Utils.notNullNotEmpty(idsRequest.getQuery())
-                || Utils.notNullNotEmpty(idsRequest.getSort());
+                || Utils.notNullNotEmpty(idsRequest.getSort())
+                || hasIsoformIds;
+    }
+
+    private boolean hasIsoformIds(List<String> ids) {
+        return ids.stream().anyMatch(id -> id.contains("-"));
```
ktibrahim, line 164: We can use a DASH constant here.

PR #333, opened 2023-03-28
Embeddings async

ahmadshadab:


common-rest/src/main/java/org/uniprot/api/rest/download/queue/EmbeddingsRabbitMQConfig.java
```diff
@@ -0,0 +1,55 @@
+package org.uniprot.api.rest.download.queue;
+
+import org.springframework.amqp.core.*;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+
+@Configuration
+@Profile({"asyncDownload"})
+public class EmbeddingsRabbitMQConfig {
```
LeonardoGonzales, line 10: I think this is not in common, because it is specific for uniprotkb

common-rest/src/main/java/org/uniprot/api/rest/download/queue/EmbeddingsQueueConfigProperties.java
```diff
@@ -0,0 +1,23 @@
+package org.uniprot.api.rest.download.queue;
+
+import lombok.Data;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@Data
+@ConfigurationProperties(prefix = "async.download.embeddings")
+public class EmbeddingsQueueConfigProperties {
```
LeonardoGonzales, line 9: Embeddings should be in UniProtKB

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/AbstractDownloadControllerIT.java
```diff
@@ -589,7 +607,9 @@ private Stream<Arguments> getSupportedFormats() {
                         APPLICATION_JSON_VALUE,
                         FASTA_MEDIA_TYPE_VALUE,
                         GFF_MEDIA_TYPE_VALUE,
-                        RDF_MEDIA_TYPE_VALUE)
+                        RDF_MEDIA_TYPE_VALUE,
+                        HDF5_MEDIA_TYPE_VALUE,
+                        "h5")
```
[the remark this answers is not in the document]
ahmadshadab: I am just testing if it works when user pass the short format name (h5) or full name HDF5_MEDIA_TYPE_VALUE (application/x-hdf5)

common-rest/src/main/java/org/uniprot/api/rest/download/queue/EmbeddingsQueueConfigProperties.java
ahmadshadab: Yes, you are right. But i kept it in common so that queue related configs are at one place rather than scattered. If you still feel we should move it to uniprotkb-rest, i can move it.

LeonardoGonzales: The problem of having it in common, is that these Bean will be loaded for every module that has asyncDownload profile (uniprot now, but idmapping, uniparc and uniref in a near feature), and it should only be instantiated for uniprotkb. Does it make sense to you?

ahmadshadab: Yes, i agree. thats a good point. Moved to uniprotkb-rest

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/AbstractDownloadControllerIT.java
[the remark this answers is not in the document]
LeonardoGonzales: I think the best place for h5 test would be at UniProtMediaTypeTest.checkMediaTypeFileExtensionRoundTrip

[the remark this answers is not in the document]
LeonardoGonzales: Maybe also add a unit test there

LeonardoGonzales (approved):

PR #334, opened 2023-04-17
add support for multiple sequence range for the same accession

ahmadshadab:


idmapping-rest/src/main/java/org/uniprot/api/idmapping/service/PIRResponseConverter.java
```diff
@@ -101,46 +101,59 @@ public IdMappingResult convertToIDMappings(
...
                             .map(this::getMappedId)
                             .filter(Objects::nonNull)
-                            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+                            .collect(
+                                    Collectors.groupingBy(
+                                            Map.Entry::getKey,
+                                            Collectors.mapping(
+                                                    Map.Entry::getValue,
+                                                    Collectors.toCollection(LinkedHashSet::new))));
+            // remove more than one versions from values
```
LeonardoGonzales, line 151: Can you please explain me why we are removing the version for just first value?

idmapping-rest/src/test/java/org/uniprot/api/idmapping/service/impl/PIRServiceImplTest.java
```diff
@@ -170,7 +167,39 @@ void testThrowsExceptionWhenDuplicateIds() {
                         "http://localhost", new HttpEntity<>(map, HTTP_HEADERS), String.class))
                 .thenReturn(ResponseEntity.ok().body("P00001\tP00001"));
 
-        Assertions.assertThrows(
-                IllegalStateException.class, () -> pirService.mapIds(request, "dummyJobId"));
+        IdMappingResult idMappingResult = pirService.mapIds(request, "dummyJobId");
+        assertThat(
+                idMappingResult.getMappedIds(),
+                contains(new IdMappingStringPair("P00001.1", "P00001")));
```
LeonardoGonzales, line 173: I tested this use case in legacy and it returns without version.
In Legacy it would return "P00001", "P00001" --> it also would return that it found different versions, but I think it is another Jira.
Let me know if you have any doubt This is my test URL for legacy: https://legacy.uniprot.org/uniprot/?query=yourlist:M20230420A084FC58F6BBA219896F365D15F2EB4454E3D78&sort=yourlist:M20230420A084FC58F6BBA219896F365D15F2EB4454E3D78&columns=yourlist(M20230420A084FC58F6BBA219896F365D15F2EB4454E3D78),id,entry%20name,reviewed,protein%20names,genes,organism,length,comment(INVOLVEMENT%20IN_DISEASE)

idmapping-rest/src/main/java/org/uniprot/api/idmapping/service/PIRResponseConverter.java
ahmadshadab: if we have more than one version of the same accession, we return just main entry with mapping.
e.g. P05067.304 -> P05067
If we don't remove version, we will have duplicate mapping and it will be different that current implementation

idmapping-rest/src/test/java/org/uniprot/api/idmapping/service/impl/PIRServiceImplTest.java
ahmadshadab: I think, legacy is not correct. We are returning correct mapping for version from beginning. 
if you map P05067.304 then we get P05067.304 -> P05067 (which is more consistent i would say)

PR #341, opened 2023-05-18
Advanced search: remove ontologies/go field from cross-references

ktibrahim:


support-data-rest/src/main/java/org/uniprot/api/support/data/configure/service/UniProtKBConfigureService.java
```diff
@@ -70,7 +70,12 @@ public List<EvidenceGroup> getGoEvidences() {
...
         List<DatabaseGroup> databases =
                 Arrays.stream(UniProtDatabaseCategory.values())
-                        .filter(dbCat -> dbCat != UniProtDatabaseCategory.UNKNOWN)
+                        .filter(
+                                dbCat ->
+                                        (dbCat != UniProtDatabaseCategory.UNKNOWN
+                                                && dbCat
+                                                        != UniProtDatabaseCategory
+                                                                .GENE_ONTOLOGY_DATABASES))
                         .map(this::getDatabaseGroup)
```
ahmadshadab, line 79: take out the filter condition to a method with meaningful name like 
boolean filteredUniProtDBCategory(dbCategory){
 // code with conditions
}

support-data-rest/src/test/java/org/uniprot/api/support/data/configure/service/UniProtKBConfigureServiceTest.java
```diff
@@ -39,7 +39,7 @@ void testGetUniProtSearchCrossReferences() {
         assertNotNull(crossRefGroup.getItems());
 
         List<AdvancedSearchTerm> crossRefCat = crossRefGroup.getItems();
-        assertEquals(21, crossRefCat.size());
+        assertEquals(20, crossRefCat.size());
         AdvancedSearchTerm anyGroup = crossRefCat.get(0);
```
ahmadshadab, line 43: verify that crossRefCat doesn't have id `xref_group_ontologies`

PR #346, opened 2023-06-06
TRM-29103: idMapping for deleted UniProtKB entries does not show status

LeonardoGonzales:
idMapping for deleted UniProtKB entries does not return inactiveStatus

### DIFF
--- a/idmapping-rest/src/test/java/org/uniprot/api/idmapping/controller/UniProtKBIdMappingStreamControllerIT.java
+++ b/idmapping-rest/src/test/java/org/uniprot/api/idmapping/controller/UniProtKBIdMappingStreamControllerIT.java
@@ -7,7 +7,6 @@ import static org.springframework.http.HttpHeaders.ACCEPT;
 import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.log;
-import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.*;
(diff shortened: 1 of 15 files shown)


idmapping-rest/src/test/java/org/uniprot/api/idmapping/controller/UniProtKBIdMappingStreamControllerIT.java
```diff
@@ -376,6 +381,36 @@ void streamCanReturnLineageData() throws Exception {
...
+                        .createAndPutJobInCache(
+                                UNIPROTKB_AC_ID_STR, UNIPROTKB_STR, "Q00001,I8FBX0");
+        ResultActions response =
+                performRequest(
+                        get(getIdMappingResultPath(), job.getJobId())
+                                .param("fields", "accession")
+                                .header(ACCEPT, APPLICATION_JSON_VALUE));
+        // then
+        response.andDo(print())
+                .andExpect(status().is(HttpStatus.OK.value()))
```
ahmadshadab, line 398: Replace print method with log

idmapping-rest/src/test/java/org/uniprot/api/idmapping/controller/utils/AbstractJobOperation.java
```diff
@@ -92,7 +97,7 @@ protected IdMappingJobRequest createRequest(String from, String to, String fromI
         return request;
     }
 
-    private String generateHash(IdMappingJobRequest request) {
+    private String generateHash() {
         return UUID.randomUUID().toString();
```
ahmadshadab, line 101: thanks

ahmadshadab (commented):
looks good to me. 1-2 minor comments. But could you please add a test to check if the accession is neither in store nor deleted(in solr)?

LeonardoGonzales:
This is not a valid use case. Because if the id is not in solr and is not in store it will be returned as unmappedId by PIR. 
Remember this is the result use case, all the ids sent were already verified by PIR.

ahmadshadab (approved):

PR #348, opened 2023-06-12
TRM-29130: In idmapping, unable to map an old TrEMBL ID to the SP

LeonardoGonzales:


ahmadshadab (commented):
Hi Leo,
How are you verifying that if the organism name is correct in old Trembl Entry?
I can pass O48791_SHADAB and the idmapping will still work.

idmapping-rest/src/main/java/org/uniprot/api/idmapping/service/impl/PIRServiceImpl.java
```diff
@@ -53,6 +53,11 @@ public class PIRServiceImpl extends IdMappingPIRService {
                                     .getSearchFieldItemByName("accession_id")
                                     .getValidRegex()
                             + "(\\[[0-9]+\\-[0-9]+\\]|\\.[0-9]+)");
+    public static final Pattern UNIPROTKB_ACCESSION_REGEX =
+            Pattern.compile(
+                    SearchFieldConfigFactory.getSearchFieldConfig(UniProtDataType.UNIPROTKB)
```
ktibrahim, line 58: You can use FieldRegexConstants.UNIPROTKB_ACCESSION_REGEX.

idmapping-rest/src/test/java/org/uniprot/api/idmapping/controller/utils/AbstractJobOperation.java
```diff
@@ -47,10 +49,11 @@ private String getValue(String value) {
         int subSequenceIndex = value.indexOf("[");
         if (subSequenceIndex > 0) {
             result = value.substring(0, subSequenceIndex);
-        } else {
-            int versionIndex = value.indexOf(".");
-            if (versionIndex > 0) {
-                result = value.substring(0, versionIndex);
+        } else if (value.indexOf(".") > 0) {
```
ktibrahim, line 52: Reuse constants

idmapping-rest/src/main/java/org/uniprot/api/idmapping/service/impl/PIRServiceImpl.java
ktibrahim: UNIPROTKB_ACCESSION_REGEX pattern is used in multiple places so it would be best to create a common function to get UNIPROTKB_ACCESSION_REGEX pattern. You can pass the regex to the function and get pattern in return.

ahmadshadab (approved):

ktibrahim (approved):

PR #349, opened 2023-06-13
avoid voldemort for list format in search and stream api calls

ahmadshadab:
In Solr, only the searchable fields are kept, while Voldemort keeps the entire entry object. Our search APIs utilize Solr to search for entries based on the provided query. The returned IDs from Solr are then used to retrieve the corresponding objects from Voldemort. As the list format provides the IDs only and we already have the IDs from Solr, there is no need to access Voldemort for the complete objects. The same applies to the streaming API, where accessing Voldemort is unnecessary.

### DIFF
--- a/common-rest/src/main/java/org/uniprot/api/rest/service/BasicSearchService.java
+++ b/common-rest/src/main/java/org/uniprot/api/rest/service/BasicSearchService.java
@@ -1,7 +1,6 @@
 package org.uniprot.api.rest.service;
 
 import static org.uniprot.api.rest.output.PredefinedAPIStatus.LEADING_WILDCARD_IGNORED;
-import static org.uniprot.api.rest.output.UniProtMediaType.LIST_MEDIA_TYPE_VALUE;
 
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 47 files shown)


ktibrahim (approved):

common-rest/src/main/java/org/uniprot/api/rest/service/BasicSearchService.java
```diff
@@ -102,7 +103,17 @@ public R getEntity(String idField, String value) {
     public QueryResult<R> search(SearchRequest request) {
         SolrRequest solrRequest = createSearchSolrRequest(request);
         QueryResult<D> results = repository.searchPage(solrRequest, request.getCursor());
-        Stream<R> converted = results.getContent().map(entryConverter).filter(Objects::nonNull);
+        Stream<R> converted;
```
LeonardoGonzales, line 106: We could extract line 105 that was in the old code into a new method, and override it in StoreStreamerSearchService. And you could also move mapToThinEntry to StoreStreamerSearchService class and make it an abstract. This way every Service that uses Voldemort (Store) must implement it.

common-rest/src/main/java/org/uniprot/api/rest/service/BasicSearchService.java
```diff
@@ -102,7 +103,17 @@ public R getEntity(String idField, String value) {
...
+        if (LIST_MEDIA_TYPE_VALUE.equals(request.getFormat())) {
+            converted =
+                    results.getContent()
+                            .map(Document::getDocumentId)
+                            .map(this::mapToThinEntry)
+                            .filter(Objects::nonNull);
+        } else {
+            converted = results.getContent().map(entryConverter).filter(Objects::nonNull);
```
LeonardoGonzales, line 114: This else block could call super method (extracted from line 105) that is defined in the BasicSearchService

LeonardoGonzales (approved):

PR #350, opened 2023-06-13
Entry version parameter support for UniProt entry

ktibrahim:


uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/UniProtKBController.java
```diff
@@ -199,12 +200,20 @@ public ResponseEntity<MessageConverterContext<UniProtKBEntry>> getByAccession(
...
+        if (Utils.notNullNotEmpty(version)
+                && ACCESSION_REGEX_PATTERN.matcher(accessionOrId).matches()) {
+            return redirectToEntryVersion(accessionOrId, request);
+        }
         if (accessionOrId.contains("_")) {
             String accession = entryService.findAccessionByProteinId(accessionOrId);
             return redirectToAccession(accessionOrId, accession, request);
         } else if (accessionOrId.contains(".")) {
-            return redirectToUniSave(accessionOrId, request);
+            return redirectToUniSave(accessionOrId, request, Optional.ofNullable(null));
```
LeonardoGonzales, line 216: Can you please use Optional.empty()

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/UniProtKBController.java
```diff
@@ -218,6 +227,30 @@ public ResponseEntity<MessageConverterContext<UniProtKBEntry>> getByAccession(
         }
     }
 
+    private ResponseEntity<MessageConverterContext<UniProtKBEntry>> redirectToEntryVersion(
+            String accessionOrId, HttpServletRequest request) {
+        String version = null;
+        if (request.getParameter("version").equals("last")) {
+            try {
+                String entryVersionHistoryURI = "https://rest.uniprot.org/unisave/" + accessionOrId;
```
LeonardoGonzales, line 235: Can you please use the logic bellow (in the method redirectToUniSave) to get UniSave path, extract a method and reuse in both places

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/UniProtKBController.java
```diff
@@ -218,6 +227,30 @@ public ResponseEntity<MessageConverterContext<UniProtKBEntry>> getByAccession(
...
+                String response = Utils.httpGetRequest(entryVersionHistoryURI);
```
LeonardoGonzales, line 236: I think we need to be prepared for an accession that does not exist. Here we are assuming that all accessions received as a parameter exists in UniSave

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/UniProtKBByAccessionControllerIT.java
```diff
@@ -566,8 +573,103 @@ void searchWithIDForDeletedInactiveEntriesRedirectToAccession() throws Exception
...
+                .andExpect(header().string(HttpHeaders.CONTENT_TYPE, TSV_MEDIA_TYPE_VALUE))
+                .andExpect(
+                        content()
+                                .string(
+                                        containsString(
+                                                "Expected one of [text/plain;format=fasta, text/plain;format=flatfile]")));
+    }
+
+    private String retrieveLastEntryVersion(String accession) throws Exception {
+        String entryVersionHistoryURI = "https://rest.uniprot.org/unisave/" + accession;
```
LeonardoGonzales, line 660: Here in the test we are accessing external resources, this may cause issues and it is not recommended.

LeonardoGonzales (commented):
I would add a test for an accession that does not exist with version parameter, because your logic in the controller is accessed before try to find the accession. Let's chat about it if you want.

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/UniProtKBController.java
ktibrahim: The redirectToUniSave method constructs the unisave URI and redirects the request to it. I think it's better to keep the logic for getting entry version separate from the unisave URI construction & redirection logic. What do you think?

The purpose of redirectToEntryVersion is to get the entry version and pass that version to unisave. Incase of "last" version we are calling unisave history endpoint to fetch all entry versions for the accession. Get the latest/last version and pass it to redirectToUniSave method.

LeonardoGonzales: The reason that we implemented that way was because this code can also be executed with another URL, for example, when we create our dev environment, we will have a different URL for it, and it will probably be something like https://wwwdev.ebi.ac.uk/uniprot/api/unisave/
If we keep hardcoded rest.uniprot.org, it will be inconsistent in our dev environment and not redirecting to the correct unisave.

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/UniProtKBController.java
LeonardoGonzales: This error response will break your code in line 240. because your code is always expecting "results" in the json, and when you receive back a json with error message, it will break

LeonardoGonzales (approved):

PR #355, opened 2023-06-30
TRM-29823: Improve Cache control to use no-cache for error (500) requests

LeonardoGonzales:


common-rest/src/main/java/org/uniprot/api/rest/output/header/HttpCommonHeaderConfig.java
```diff
@@ -116,7 +119,9 @@ void handleGatewayCaching(HttpServletRequest request, HttpServletResponse respon
                 break;
             }
         }
-        if (requiresCachingHeaders) {
+        if (requiresCachingHeaders
+                && (HttpServletResponse.SC_OK == response.getStatus()
+                        || HttpServletResponse.SC_SEE_OTHER == response.getStatus())) {
             // request gateway caching
```
ahmadshadab, line 125: Please filter specific codes which don't need cache

ahmadshadab (commented):
Please merge after adding that change

common-rest/src/main/java/org/uniprot/api/rest/output/header/HttpCommonHeaderConfig.java
LeonardoGonzales: I updated to < 400, as we discussed in our chat

ahmadshadab (approved):

PR #357, opened 2023-07-05
TRM-29849: Error when downloading ID mapping results

LeonardoGonzales:


idmapping-rest/src/test/java/org/uniprot/api/idmapping/controller/UniProtKBIdMappingStreamControllerIT.java
```diff
@@ -527,4 +565,18 @@ void testIdMappingWithSubSequenceRepeatedAccessionValid() throws Exception {
                 .andExpect(content().string(containsString(">tr|Q00001|20-30\nLARPSFSLVED\n")))
                 .andExpect(content().string(containsString(">sp|Q00002|20-30\nLARPSFSLVED")));
     }
+
```
ahmadshadab, line 568: Please add test for XLS format as well.

ahmadshadab (commented):
When the user passed Inactive entry id, what will be the impact on the following APIs after putting the null check in the converter? Will this change not change the existing behaviour?

1. get by id
2. get by ids(accessions)
3. search
4. stream

idmapping-rest/src/test/java/org/uniprot/api/idmapping/controller/UniProtKBIdMappingStreamControllerIT.java
LeonardoGonzales: Done

LeonardoGonzales:
For the other APIs I created another JIRA: TRM-29856

PR #358, opened 2023-07-06
Hot fix 1 0 25 2

ahmadshadab:


pom.xml
```diff
@@ -12,7 +12,7 @@
 
 	<artifactId>uniprot-rest-api</artifactId>
 	<groupId>org.uniprot</groupId>
-	<version>1.0.26-SNAPSHOT</version>
```
supun-ebi, line 15: we can declare the version as a property in the main pom (here) and reuse it elsewhere.

jacoco-aggregate-report/pom.xml
```diff
@@ -5,7 +5,7 @@
 	<parent>
 		<artifactId>uniprot-rest-api</artifactId>
 		<groupId>org.uniprot</groupId>
-		<version>1.0.26-SNAPSHOT</version>
+		<version>1.0.25.2</version>
```
supun-ebi, line 8: Can't we use ${project.parent.version} for these places as well (including other similar places here)?

LeonardoGonzales: This is updated by release script. I updated back to 1.0.26-SNAPSHOT, this way we can merge it to master

support-data-common/src/main/resources/keyword-query.config
ahmadshadab, line 1: Leo, we should remove the following from support-data-rest, since they are already there in support-data-common

1. keyword-query.conf
2. taxonomy-query.config

support-data-common/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>support-data-common</artifactId>
+	<description>UniProt support data lib</description>
+
+	<properties>
+		<maven.compiler.source>11</maven.compiler.source>
+		<maven.compiler.target>11</maven.compiler.target>
+	</properties>
```
ahmadshadab, line 18: Do we need java version? I don't see in other module

ahmadshadab (commented):
Please make the changes in k8s meta-cm in a private branch. We can merge in master before hotfix.

support-data-common/src/main/resources/keyword-query.config
LeonardoGonzales: As you can see in the header of the comment, keyword-query.config is already in support-data-common module.

support-data-common/pom.xml
LeonardoGonzales: thank you, removing it

support-data-common/src/main/resources/keyword-query.config
ahmadshadab: I didn't get you. What i am saying is that the files keyword-query.conf and taxonomy-query.config are there in two modules(support-data-common and support-data-rest). They should only be in module support-data-common.

supun-ebi: I also agree with the fact that those 2 can be removed.

LeonardoGonzales: Sorry, I got confused with facet config files. I removed it from rest and executed the build to make sure that it would not impact the build. It should be fine now.

support-data-common/pom.xml
LeonardoGonzales: done

supun-ebi (approved):

PR #359, opened 2023-07-12
Trm 29852 filter isoforms

supun-ebi:


uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/service/UniProtEntryServiceTest.java
```diff
@@ -214,6 +220,30 @@ void stream_in_list_format_without_voldemort_being_called() {
         assertEquals(entry2, entries.get(1));
     }
 
+    @Test
+    void getFacets() {
```
LeonardoGonzales, line 224: Can you please add another test when we do not need to add filterQuery?

LeonardoGonzales (approved):
Thank you! After add the test you can merge it

PR #361, opened 2023-07-12
add parent info

supun-ebi:


uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/groupby/service/GroupByECService.java
```diff
@@ -11,6 +8,9 @@
 import org.uniprot.api.uniprotkb.service.UniProtEntryService;
 import org.uniprot.core.cv.ec.ECEntry;
 
+import java.util.*;
```
LeonardoGonzales, line 11: Can you please run maven command to make sure that import order is based on maven plugin?

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/groupby/service/GroupByECServiceTest.java
```diff
@@ -28,6 +16,18 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.empty;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+import static org.uniprot.api.uniprotkb.groupby.service.GroupByECService.*;
+
 @ExtendWith(MockitoExtension.class)
 class GroupByECServiceTest {
```
LeonardoGonzales, line 32: Excellent, the rename!!

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/groupby/service/go/GOServiceTest.java
```diff
@@ -27,11 +29,21 @@ void setUp() {}
...
 
         List<GoRelation> children = goService.getChildren(ID);
 
         verify(goClient).getChildren(ID);
-        assertSame(result, children);
+        assertSame(resultChildren, children);
+    }
+
+    @Test
+    void getGoRelation() {
```
LeonardoGonzales, line 41: Do you think a test for not found should be added? 
Do you think the same would also apply for getChildren?

supun-ebi: I don't think we need separate tests since empty list and empty optional scenarios are covered in the written tests. (mocked results can get one of any form)

PR #363, opened 2023-07-20
Test & verify cc_similarity field for Tsv

ktibrahim:


common-rest/src/test/java/org/uniprot/api/rest/output/converter/TsvMessageConverterTest.java
```diff
@@ -66,7 +66,8 @@ void canWriteBody() throws IOException {
         assertNotNull(result);
         assertTrue(
                 result.contains(
-                        "P00001\t"
+                        "Entry\tSite\tFunction [CC]\tSequence similarities\n"+
+                                "P00001\t"
```
LeonardoGonzales, line 70: We should write the content of Sequence similarities as well. Here we are just adding the header. There is something wrong.

PR #366, opened 2023-07-24
Trm 29855 groupby go error

supun-ebi:


uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/groupby/service/go/client/GOClientImpl.java
```diff
@@ -29,7 +28,9 @@ public List<GoRelation> getChildren(String goId) {
 
         String url = goApiPrefix + goId + "/children";
         GoTermResult result = restTemplate.getForObject(url, GoTermResult.class);
-        if (result == null || result.getResults().isEmpty()) {
+        if (result == null
+                || result.getResults().isEmpty()
+                || result.getResults().get(0).getChildren() == null) {
```
LeonardoGonzales, line 33: Do you think we should cover it in a test?

PR #372, opened 2023-08-04
return obsolete count for idmapping

ahmadshadab:


common-rest/src/main/java/org/uniprot/api/rest/output/converter/JsonMessageConverter.java
```diff
@@ -155,6 +155,11 @@ protected void after(MessageConverterContext<T> context, OutputStream outputStre
                 if (notNullNotEmpty(extraOptions.getSuggestedIds())) {
                     writeJsonArray(generator, "suggestedIds", extraOptions.getSuggestedIds());
                 }
+
+                if (Objects.nonNull(extraOptions.getObsoleteCount())) {
+                    generator.writeFieldName("obsoleteCount");
```
supun-ebi, line 160: what if we create a method for this like 'writeJsonObject'? (similar to writeJsonArray)

idmapping-rest/src/test/java/org/uniprot/api/idmapping/controller/AbstractIdMappingResultsControllerIT.java
```diff
@@ -13,6 +13,7 @@
 import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.log;
+import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
```
supun-ebi, line 16: unused import

supun-ebi (approved):

PR #374, opened 2023-08-17
add validator on fields param on async download run api

ahmadshadab:


common-rest/src/main/java/org/uniprot/api/rest/validation/CustomConstraintGroup.java
```diff
@@ -0,0 +1,3 @@
+package org.uniprot.api.rest.validation;
+
+public interface CustomConstraintGroup {}
```
supun-ebi, line 3: Can we have a more specific name that shows the purpose? 
Ex: DownloadSubmissionVerificationGroup

common-rest/src/main/java/org/uniprot/api/rest/validation/CustomConstraintGroupSequence.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * When validating CustomConstraintGroupSequence, first the Default group is validated. If all the
+ * data passes validation, then the CustomConstraintGroup group is validated. If a constraint is
+ * part of both the Default and the CustomConstraintGroup groups, the constraint is validated as
+ * part of the Default group and will not be validated on the subsequent CustomConstraintGroup pass.
+ * To use on an object @Validated(CustomConstraintGroupSequence.class)
+ */
+@GroupSequence({Default.class, CustomConstraintGroup.class})
+public interface CustomConstraintGroupSequence {}
```
supun-ebi, line 14: Same as above

common-rest/src/main/java/org/uniprot/api/rest/validation/ValidDownloadRequest.java
```diff
@@ -0,0 +1,121 @@
...
+@Target({ElementType.TYPE, ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ValidDownloadRequest {
+
+    /** formats where 'fields' param doesn't make sense and hence not allowed */
+    List<String> FORMATS_WITH_NO_PROJECTION =
+            List.of(
+                    FF_MEDIA_TYPE_VALUE,
+                    LIST_MEDIA_TYPE_VALUE,
+                    MediaType.APPLICATION_XML_VALUE,
```
supun-ebi, line 38: I would prefer this to be supplied as a static import as well. Just for appearance

common-rest/src/main/java/org/uniprot/api/rest/validation/ValidDownloadRequest.java
```diff
@@ -0,0 +1,121 @@
...
+
+                    String type = getMediaTypeFromShortName(format);
+
+                    if (FORMATS_WITH_NO_PROJECTION.contains(format)
+                            || FORMATS_WITH_NO_PROJECTION.contains(type)) {
+                        buildInvalidParamFieldsErrorMessage(format, contextImpl);
+                        isValid = false;
+                    }
+
+                    if (!isValid && context != null) {
```
supun-ebi, line 84: context != null is already true by now?

common-rest/src/main/java/org/uniprot/api/rest/validation/ValidDownloadRequest.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        @Override
+        public boolean isValid(
+                DownloadRequest downloadRequest, ConstraintValidatorContext context) {
+            boolean isValid = true;
+            try {
+                String fields = BeanUtils.getProperty(downloadRequest, "fields");
+                if (Utils.notNullNotEmpty(fields)) {
+                    ConstraintValidatorContextImpl contextImpl =
```
supun-ebi, line 71: contextImpl can be defined inside the next if block?

common-rest/src/main/java/org/uniprot/api/rest/validation/ValidDownloadRequest.java
```diff
@@ -0,0 +1,121 @@
...
+                            (ConstraintValidatorContextImpl) context;
+
+                    String format = getPassedOrDefaultFormat(downloadRequest);
```
supun-ebi, line 74: I would prefer these to be added into a separate method for improved readability

private boolean isPassedWithInvalidFormat(DownloadRequest downloadRequest){
    String format = getPassedOrDefaultFormat(downloadRequest);
    String type = getMediaTypeFromShortName(format);

     return FORMATS_WITH_NO_PROJECTION.contains(format) || FORMATS_WITH_NO_PROJECTION.contains(type)
 }
 
 then add the if check using the method

common-rest/src/test/java/org/uniprot/api/rest/validation/ValidDownloadRequestTest.java
```diff
@@ -0,0 +1,64 @@
...
+    void when_fields_passed_with_short_format_then_invalid_request() {
+        FakeDownloadRequest downloadRequest = new FakeDownloadRequest();
+        downloadRequest.setFormat("h5");
+        downloadRequest.setFields("field1,field2");
+        FakeDownloadRequestValidator validator = new FakeDownloadRequestValidator();
+        boolean result = validator.isValid(downloadRequest, null);
+        assertFalse(result);
+        assertNotNull(validator.errorMessage);
+    }
+
```
supun-ebi, line 54: Don't we need cases for when a valid format is passed?

common-rest/src/test/java/org/uniprot/api/rest/validation/ValidDownloadRequestTest.java
```diff
@@ -0,0 +1,64 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorContextImpl;
+import org.junit.jupiter.api.Test;
+import org.uniprot.api.rest.request.FakeDownloadRequest;
+
+class ValidDownloadRequestTest {
+
+    @Test
+    void when_fields_passed_then_invalid_request() {
```
supun-ebi, line 12: isn't it when fields passed with an invalid (not accepted) format?

common-rest/src/main/java/org/uniprot/api/rest/validation/CustomConstraintGroup.java
ahmadshadab: The reason for choosing that name was that the Group can be used for any other validation where we need custom behaviour.

common-rest/src/main/java/org/uniprot/api/rest/validation/ValidDownloadRequest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.uniprot.api.rest.request.DownloadRequest;
+import org.uniprot.core.util.Utils;
+
+@Constraint(validatedBy = ValidDownloadRequest.DownloadRequestValidator.class)
+@Target({ElementType.TYPE, ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ValidDownloadRequest {
+
+    /** formats where 'fields' param doesn't make sense and hence not allowed */
+    List<String> FORMATS_WITH_NO_PROJECTION =
```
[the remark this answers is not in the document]
ahmadshadab: I will make this change. But for an input size of 9, how much it matters?

common-rest/src/test/java/org/uniprot/api/rest/validation/ValidDownloadRequestTest.java
ahmadshadab: yes. i will add. thanks

common-rest/src/test/java/org/uniprot/api/rest/validation/ValidDownloadRequestTest.java
ahmadshadab: yes, renaming the method name

common-rest/src/main/java/org/uniprot/api/rest/validation/ValidDownloadRequest.java
ahmadshadab: from test the context will be null.

common-rest/src/main/java/org/uniprot/api/rest/validation/ValidDownloadRequest.java
```diff
@@ -105,13 +108,18 @@ private static String getMediaTypeFromShortName(String format) {
...
                 throws InvocationTargetException, IllegalAccessException, NoSuchMethodException {
-            String format = BeanUtils.getProperty(downloadRequest, "format");
+            String format = getPassedFormat(downloadRequest);
             if (Utils.nullOrEmpty(format)) {
                 format = APPLICATION_JSON_VALUE;
             }
             return format;
         }
 
+        private String getPassedFormat(DownloadRequest downloadRequest)
```
supun-ebi, line 118: You may change this method to be reused in line 69.

common-rest/src/test/java/org/uniprot/api/rest/validation/ValidDownloadRequestTest.java
```diff
@@ -9,9 +9,21 @@
...
+        boolean result = validator.isValid(downloadRequest, null);
+        assertTrue(result);
+        assertNull(validator.errorMessage);
+    }
+
+    @Test
+    void when_fields_passed_with_unsupported_format_then_invalid_request() {
+        FakeDownloadRequest downloadRequest = new FakeDownloadRequest();
+        downloadRequest.setFormat(
+                ValidDownloadRequest.FORMATS_WITH_NO_PROJECTION.stream().findFirst().get());
```
supun-ebi, line 26: iterator().next() could be a simpler way here.

ahmadshadab: agree.

common-rest/src/main/java/org/uniprot/api/rest/validation/ValidDownloadRequest.java
```diff
@@ -68,16 +68,11 @@ public boolean isValid(
...
-                    if (FORMATS_WITH_NO_PROJECTION.contains(format)
-                            || FORMATS_WITH_NO_PROJECTION.contains(type)) {
-                        buildInvalidParamFieldsErrorMessage(format, contextImpl);
+                    if (isPassedWithInvalidFormat(downloadRequest)) {
+                        ConstraintValidatorContextImpl contextImpl =
+                                (ConstraintValidatorContextImpl) context;
+                        buildInvalidParamFieldsErrorMessage(
+                                getPassedFormat(downloadRequest), contextImpl);
                         isValid = false;
                     }
```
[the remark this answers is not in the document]
ahmadshadab: good point. it was copy paste from other validator. :-)

supun-ebi (approved):

PR #385, opened 2023-09-19
Add brotli compression for voldemort

ahmadshadab:
Please see Jira  [TRM-30169](https://www.ebi.ac.uk/panda/jira/browse/TRM-30169) for details


LeonardoGonzales (approved):

LeonardoGonzales (commented):
Hi Shadab, 
In the client we need to read new added properties as well, We need to use at least the enabledBrotli (The compression quality is not necessary in the read). 
Check the class (UniProtStoreConfig), it also need to be checked for idmapping, uniref and uniparc code.

LeonardoGonzales (approved):

supun-ebi (approved):

PR #392, opened 2023-11-08
Trm 30200 upgrade status endpoint

supun-ebi:


common-rest/src/main/java/org/uniprot/api/rest/download/queue/AbstractMessageListener.java
```diff
@@ -64,30 +64,33 @@ protected void processMessage(Message message, DownloadJob downloadJob) {
...
 
-    protected void writeSolrResult(DownloadRequest request, Path idsFile, String jobId)
+    protected void writeSolrResult(DownloadRequest request, DownloadJob downloadJob, Path idsFile)
             throws IOException {
-        Stream<String> ids = streamIds(request);
-        writeIdentifiers(idsFile, ids);
-        log.info("Solr ids saved for job {}", jobId);
+        writeIdentifiers(idsFile, streamIds(request));
+        updateTotalEntries(downloadJob, getNumberOfEntries(request));
+        log.info("Solr ids saved for job {}", downloadJob.getId());
```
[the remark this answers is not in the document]
ahmadshadab: I didn't see integration test for processedEntries update. It would be nice to have a test to verify increment of processedEntries count

idmapping-rest/src/main/java/org/uniprot/api/idmapping/queue/AbstractIdMappingDownloadResultWriter.java
```diff
@@ -122,6 +136,18 @@ public void writeResult(
...
+        downloadJob.setUpdated(LocalDateTime.now());
+        jobRepository.save(downloadJob);
+    }
+
+    private void updatedNoOfProcessedEntriesByBatch(DownloadJob downloadJob, int size) {
+        updatedNoOfProcessedEntries(
+                downloadJob,
+                Math.min(size, downloadJob.getTotalEntries() - downloadJob.getEntriesProcessed()));
+    }
+
```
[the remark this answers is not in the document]
supun-ebi: It's very hard to do so since we only treat the response as a list of strings.

common-rest/src/main/java/org/uniprot/api/rest/download/queue/AbstractMessageListener.java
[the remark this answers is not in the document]
supun-ebi: (1) Tagging @LeonardoGonzales to get an idea
(2) I have updated tests on AbstractAsyncDownloadIT and XIdMappingDownloadResultWriterTest classes. In fact the coverage whole async download is not so great, and I didnt add any new test classes. (Could be a sub task in refactoring async download code task)

common-rest/src/main/java/org/uniprot/api/common/repository/stream/rdf/RdfEntryCountProvider.java
```diff
@@ -0,0 +1,50 @@
...
+                                            "<http://purl.uniprot.org/uniprot/UPI[\\w]{10}> <http://purl.uniprot.org/core/reviewed>"),
+                    "uniref",
+                            Map.of(
+                                    "rdf",
+                                            "<rdf:Description rdf:about=\"+(UniRef100|UniRef90|UniRef50)_([OPQ][0-9][A-Z0-9]{3}[0-9]|[A-NR-Z]([0-9][A-Z][A-Z0-9]{2}){1,2}[0-9])(-[0-9]+)?+\">",
+                                    "ttl",
+                                            "<(UniRef100|UniRef90|UniRef50)_([OPQ][0-9][A-Z0-9]{3}[0-9]|[A-NR-Z]([0-9][A-Z][A-Z0-9]{2}){1,2}[0-9])(-[0-9]+)?> rdf:type up:Protein ;",
+                                    "nt",
+                                            "<http://purl.uniprot.org/uniprot/(UniRef100|UniRef90|UniRef50)_([OPQ][0-9][A-Z0-9]{3}[0-9]|[A-NR-Z]([0-9][A-Z][A-Z0-9]{2}){1,2}[0-9])(-[0-9]+)?> <http://purl.uniprot.org/core/reviewed>"));
+
```
ahmadshadab, line 34: Wow. Map of map. 👍 But i see a problem with regex matching. When you call matches method in isAMatch, Java compiles the regex and which is kind of heavy operation. If you are calling the same regex multiple times and Java compiles each time, it is inefficient.
Change the value of Map to Pattern instead of String and call pattern.matcher().matches(str) instead of str.matches(regex). Search `Pattern.compile` in our codebase for example.

ahmadshadab: Feel free to file a separate ticket to move all the regex at one common place. It would be easy maintain if we have one copy of regex for each data type.

common-rest/src/main/java/org/uniprot/api/common/repository/stream/rdf/RdfStreamer.java
```diff
@@ -47,6 +47,44 @@ public Stream<String> stream(Stream<String> entryIds, String dataType, String fo
...
+
+        Stream<String> rdfStringStream =
+                StreamSupport.stream(batchRDFXMLStoreIterable.spliterator(), false)
+                        .flatMap(Collection::stream)
+                        .map(
+                                response -> {
+                                    consumer.accept(
+                                            rdfEntryCountProvider.getEntryCount(
+                                                    response, dataType, format));
+                                    return response;
```
ahmadshadab, line 66: The name response gives an impression that it is full response but it is just a line of the big response. 
line would be more clearer for the reader.

common-rest/src/main/java/org/uniprot/api/rest/download/AbstractDownloadResultWriter.java
```diff
@@ -91,17 +98,32 @@ public void writeResult(
...
                         StreamSupport.stream(batchStoreIterable.spliterator(), false)
+                                .map(
+                                        entityCollection -> {
+                                            updateEntriesProcessed(
+                                                    downloadJob, entityCollection.size());
+                                            return entityCollection;
+                                        })
                                 .flatMap(Collection::stream)
                                 .onClose(
                                         () ->
```
ahmadshadab, line 129: Put a try catch in the code where we are updating the processed count and continue with the remaining flow.

ahmadshadab (approved):
Please do not merge until the other task to introduce two configs is done.
Just handle the below two comments and treat it as approved:

1. IntConsumer to LongConsumer
2. Try catch during process count update.

PR #394, opened 2023-11-16
allow more than one servers in cross ref

ahmadshadab:


common-rest/src/main/java/org/uniprot/api/rest/download/queue/BaseAbstractMessageListener.java
```diff
@@ -124,9 +126,12 @@ protected static boolean isJobSeenBefore(
                 && downloadJob.getStatus() != JobStatus.ERROR;
     }
 
-    protected void writeIdentifiers(Path filePath, Stream<String> ids) throws IOException {
-        Iterable<String> source = ids::iterator;
-        Files.write(filePath, source, StandardOpenOption.CREATE);
+    protected int writeIdentifiers(Path filePath, Stream<String> ids) throws IOException {
```
LeonardoGonzales, line 129: Can you please rollback this change, because it is not part of this task?

LeonardoGonzales (approved):

PR #412, opened 2024-01-18
Split UniProtKB rest into rest and common

ahmadshadab:


LeonardoGonzales:
Currently we have some uniprotkb related classes that are in common-rest (common to all rest projects) and they are good candidates to be moved to uniprotkb-common (because they are used in uniprotkb functionalities for idmapping results as well), for example, TaxonomyLineage related classes.

LeonardoGonzales (commented):
I tried to run sonar review but it failed. I think we need to merge main branch into this branch again and try to run it.

common-rest/src/test/java/org/uniprot/api/rest/controller/AbstractSearchControllerIT.java
```diff
@@ -551,7 +551,9 @@ void searchFieldsWithIncorrectValuesReturnBadRequest() throws Exception {
     @MethodSource("getAllReturnedFields")
     void searchCanSearchWithAllAvailableReturnedFields(String name, List<String> paths)
             throws Exception {
-
+        if ("error_gmodel_pred".equals(name)) {
```
LeonardoGonzales, line 554: Can you please remove it?

LeonardoGonzales (commented):
We also need to do the package changes that we chatted about

ahmadshadab:
> I tried to run sonar review but it failed. I think we need to merge main branch into this branch again and try to run it.

I ran sonar after updating with main. It passed.

PR #422, opened 2024-01-26
Trm 30744 split uniref

supun-ebi:


uniref-common/pom.xml
```diff
@@ -0,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.uniprot</groupId>
+		<artifactId>uniprot-rest-api</artifactId>
+		<version>1.0.28-SNAPSHOT</version>
```
LeonardoGonzales, line 9: Can you please update it to 1.0.29-SNAPSHOT?

uniref-common/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+	</parent>
+	<artifactId>uniref-common</artifactId>
+
+	<properties>
+		<maven.compiler.source>11</maven.compiler.source>
```
LeonardoGonzales, line 14: I think we do not need to have compile properties here, it will get from parent pom...  we define it there

LeonardoGonzales:
more 1 small details.
- move MessageConverterConfig one level up in the package structure

LeonardoGonzales (approved):

PR #424, opened 2024-01-29
Trm 30746 split uniparc

supun-ebi:


LeonardoGonzales:
In uniparc-rest, we need to delete repository folder

PR #429, opened 2024-02-01
TRM-30467: update unirule_id json path and improve validation regex

LeonardoGonzales:


aa-rest/src/test/java/org/uniprot/api/aa/controller/ArbaGetByIdControllerIT.java
```diff
@@ -148,7 +148,7 @@ public GetIdParameter withFilterFieldsParameter() {
                     .id(ARBA_ID)
                     .fields("annotation_covered")
                     .resultMatcher(jsonPath("$.uniRuleId", is(ARBA_ID)))
-                    .resultMatcher(jsonPath("$.information").doesNotExist())
+                    .resultMatcher(jsonPath("$.information").exists())
```
ahmadshadab, line 151: We should also check if information.oldRuleNum exists and preferably with actual value.

ahmadshadab (approved):

PR #431, opened 2024-02-06
Javax2jakartajaxb

ahmadshadab:


common-rest/src/main/java/org/uniprot/api/rest/output/converter/AbstractUUWHttpMessageConverter.java
```diff
@@ -237,7 +237,7 @@ private void logStats(int counter, Instant start) {
         long millisDuration = Duration.between(start, now).toMillis();
         int secDuration = (int) millisDuration / 1000;
         String rate = String.format("%.2f", ((double) counter) / secDuration);
-        log.info(
+        log.debug(
```
LeonardoGonzales, line 240: Can you please return it to info?

LeonardoGonzales (approved):

PR #440, opened 2024-02-21
clean up before retry failed job

supun-ebi:


ahmadshadab (commented):
As per our slack call, please remove deletion code from other catch blocks. And also handle max retry situation.

common-rest/src/main/java/org/uniprot/api/rest/download/queue/BaseAbstractMessageListener.java
```diff
@@ -65,13 +70,18 @@ public void onMessage(Message message) {
                         "Message with job id {} discarded after max retry {}",
                         jobId,
                         getMaxRetryCount());
+                asyncDownloadFileHandler.deleteAllFiles(jobId);
                 return;
```
ahmadshadab, line 74: I think, we should reset the counts in if block after maximum retry. You are just cleaning up the files.
I feel, you should have called cleanBeforeRetry(jobId); here too instead of asyncDownloadFileHandler.deleteAllFiles(jobdi);

If you make this change, you need to rename method cleanBeforeRetry

common-rest/src/main/java/org/uniprot/api/rest/download/queue/BaseAbstractMessageListener.java
```diff
@@ -98,6 +108,16 @@ public void onMessage(Message message) {
         }
     }
 
+    private void cleanBeforeRetry(String jobId) {
+        jobRepository.update(
```
ahmadshadab, line 112: We could put private method down with other private methods.

ahmadshadab (approved):
Merge after addressing one comment about resetting counts after max retry.
Nice to have one test case which covers max retry scenario.

PR #447, opened 2024-03-05
fix tests

supun-ebi:


LeonardoGonzales (approved):
did you double check that existence:2 is an unmatched query filter?

PR #449, opened 2024-03-07
Heartbeat retry main

supun-ebi:


common-rest/src/main/java/org/uniprot/api/rest/download/heartbeat/HeartBeatProducer.java
```diff
@@ -27,6 +31,11 @@ public class HeartBeatProducer {
     private final Map<String, Long> lastSavedPoints = new HashMap<>();
     private final AsyncDownloadHeartBeatConfiguration asyncDownloadHeartBeatConfiguration;
     private final DownloadJobRepository jobRepository;
+    private final RetryPolicy<Object> retryPolicy =
+            new RetryPolicy<>()
+                    .handle(Exception.class)
+                    .withDelay(Duration.ofMillis(10000))
```
LeonardoGonzales, line 37: Can you please add the retry delay value as a property?

LeonardoGonzales (commented):
I think we can close this one  because this change is already in the other PR, right?

supun-ebi:
there's another PR for this
https://github.com/ebi-uniprot/uniprot-rest-api/pull/452

PR #451, opened 2024-03-14
fix conversion logic

supun-ebi:


uniprotkb-common/src/test/java/org/uniprot/api/uniprotkb/common/response/converter/UniProtKBGffMessageConverterTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+        uniProtKBEntryUniProtKBGffMessageConverter.before(messageContext, outputStream);
+
+        String result = outputStream.toString(StandardCharsets.UTF_8);
+        assertNotNull(result);
+        assertEquals(GFF_HEADER, result.strip());
+    }
+
+    @Test
+    void writeEntity() throws Exception {
```
LeonardoGonzales, line 34: - Can you please add another test for empty, (without features)?

LeonardoGonzales (approved):

PR #453, opened 2024-03-18
Alias accession uniparc

supun-ebi:


common-rest/src/main/java/org/uniprot/api/rest/service/query/processor/UniProtFieldQueryNodeProcessor.java
```diff
@@ -86,20 +86,21 @@ public CharSequence toQueryString(EscapeQuerySyntax escaper) {
...
-                if (searchFieldItemByAlias.isPresent()) {
-                    return searchFieldItemByAlias.get().getFieldName()
-                            + SOLR_FIELD_SEPARATOR
-                            + text;
-                }
+            } else if (UNIPROTKB_ACCESSION_FIELD.equals(field)) {
+                text = text.toUpperCase();
             }
+
+            // is an alias => return the field itself
```
LeonardoGonzales, line 97: Do you think we could extract this alias logic to a method?

supun-ebi: i dnt think extracting to a different method would make sense here. but I have simplified the flow of the current method

PR #455, opened 2024-03-21
test fix

supun-ebi:


uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/UniProtKBSearchControllerIT.java
```diff
@@ -2565,16 +2565,6 @@ protected SearchContentTypeParam searchSuccessContentTypesParam() {
                                     .resultMatcher(
                                             MockMvcResultMatchers.content()
                                                     .string(containsString("##gff-version 3")))
```
LeonardoGonzales, line 2567: Can you please also validate some GFF body text?

LeonardoGonzales (approved):

PR #460, opened 2024-04-15
changed xrefs

supun-ebi:
X ref changes from core


LeonardoGonzales (approved):

support-data-rest/src/test/java/org/uniprot/api/support/data/configure/controller/IdMappingConfigureControllerIT.java
```diff
@@ -146,7 +145,7 @@ void canGetIdMappingFields() throws Exception {
                 .andExpect(
                         jsonPath(
                                 "$.groups.[?(@.groupName=='2D gel databases')].items.*",
-                                iterableWithSize(1)))
+                                iterableWithSize(0)))
```
LeonardoGonzales, line 148: I think we can remove this group validation, because it no longer exists

PR #463, opened 2024-04-24
revert jakarta

ahmadshadab:


common-rest/src/main/java/org/uniprot/api/rest/output/converter/AbstractUUWHttpMessageConverter.java
```diff
@@ -237,7 +237,7 @@ private void logStats(int counter, Instant start) {
         long millisDuration = Duration.between(start, now).toMillis();
         int secDuration = (int) millisDuration / 1000;
         String rate = String.format("%.2f", ((double) counter) / secDuration);
-        log.debug(
+        log.info(
```
LeonardoGonzales, line 240: We can keep it debug, what do you think?

common-rest/src/main/java/org/uniprot/api/rest/output/converter/ErrorMessageXMLConverter.java
```diff
@@ -4,27 +4,26 @@
...
-import jakarta.xml.bind.Marshaller;
-
 /** @author lgonzales */
 public class ErrorMessageXMLConverter extends AbstractGenericHttpMessageConverter<ErrorInfo> {
 
     private final Marshaller marshaller;
 
     public ErrorMessageXMLConverter() {
-        super(MediaType.APPLICATION_XML, UniProtMediaType.RDF_MEDIA_TYPE);
+        super(MediaType.APPLICATION_XML);
```
LeonardoGonzales, line 26: For RDF, why do you think we should remove it?

LeonardoGonzales (approved):
I have 2 small details, the RDF one I am not sure.

common-rest/src/main/java/org/uniprot/api/rest/output/converter/AbstractUUWHttpMessageConverter.java
ahmadshadab: Yes, changed it to debug

common-rest/src/main/java/org/uniprot/api/rest/output/converter/ErrorMessageXMLConverter.java
ahmadshadab: Added it back

PR #465, opened 2024-04-26
TRM-31170 improve sort in idmapping

ahmadshadab:


LeonardoGonzales (approved):
Did you run a test with the 92K use case to check how long it is taking?

PR #474, opened 2024-06-10
support inactive entry in search of taxonomy

ahmadshadab:


support-data-rest/src/test/java/org/uniprot/api/support/data/taxonomy/controller/TaxonomySearchControllerIT.java
```diff
@@ -175,13 +222,15 @@ protected SearchParameter searchCanReturnNotFoundParameter() {
         protected SearchParameter searchAllowWildcardQueryAllDocumentsParameter() {
             return SearchParameter.builder()
                     .queryParam("query", Collections.singletonList("scientific:*"))
-                    .resultMatcher(jsonPath("$.results.*.taxonId", contains(10, 15)))
+                    .resultMatcher(jsonPath("$.results.*.taxonId", contains(10, 20, 15)))
                     .resultMatcher(
                             jsonPath(
                                     "$.results.*.scientificName",
-                                    contains("scientific10", "scientific15")))
+                                    contains("scientific10", "scientific20", "scientific15")))
```
LeonardoGonzales, line 229: Inactive entry id 15 should not have scientific name

support-data-rest/src/test/java/org/uniprot/api/support/data/taxonomy/controller/TaxonomyITUtils.java
```diff
@@ -44,7 +44,7 @@ public static TaxonomyDocument createSolrDoc(long taxId, boolean facet) {
                         .strainsAdd(new TaxonomyStrainBuilder().name("str name").build())
                         .hostsAdd(new TaxonomyBuilder().taxonId(taxId + 2).build())
                         .linksAdd("link value")
-                        .active(true)
+                        .active(facet)
```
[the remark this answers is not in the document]
ahmadshadab: I agree but this is how we are creating the inactive entry even before my change. Because for search, inactive=false is relevant rest of the fields don't matter.
See code
`@Override
    protected void saveEntry(SaveScenario saveContext) {
        saveEntry(10, true);
        saveEntry(15, true);
        saveEntry(20, false);
    }`
and even test ``defaultSearchCanFindOtherName`` which doesn't return inactive before this change.
If you want i can write a separate method to create inactive entry.

support-data-rest/src/test/java/org/uniprot/api/support/data/taxonomy/controller/TaxonomySearchControllerIT.java
ahmadshadab: I know, i was just reusing the existing method. We have been creating inactive entry with all these fields(e.g. scientific) set because of common method. This change just exposed what we are saving in solr.
If you want i can create a separate method to create inactive entry but i don't see much value to it as long as solr cares about active=true boost.

LeonardoGonzales (approved):

PR #478, opened 2024-06-26
TRM-28167: UniRef created date search and return field to updated 

LeonardoGonzales:


idmapping-rest/src/test/java/org/uniprot/api/idmapping/controller/UniRefIdMappingResultsControllerIT.java
```diff
@@ -253,7 +253,7 @@ void testGetResultsInTSV() throws Exception {
                         content()
                                 .string(
                                         containsString(
-                                                "From\tCluster ID\tCluster Name\tCommon taxon\tSize\tDate of creation")))
+                                                "From\tCluster ID\tCluster Name\tCommon taxon\tSize\tDate Modified")))
                 .andExpect(
```
ahmadshadab, line 257: In UniProtKB, we call similar field "Date of last modification" in header. We should make it consistent too.

ahmadshadab (approved):

PR #479, opened 2024-06-27
Trm 31376 uniparc async download

supun-ebi:


async-download-rest/src/test/java/org/uniprot/api/async/download/messaging/consumer/processor/uniparc/UniRefRequestProcessorTest.java
```diff
@@ -0,0 +1,42 @@
+package org.uniprot.api.async.download.messaging.consumer.processor.uniparc;
```
LeonardoGonzales, line 1: Can you please rename class file name to UniParcRequestProcessorTest

PR #502, opened 2024-09-20
code and tests

supun-ebi:


uniparc-rest/src/test/java/org/uniprot/api/uniparc/controller/UniParcStreamControllerIT.java
```diff
@@ -356,6 +554,23 @@ void streamAllContentType(MediaType mediaType) throws Exception {
                 .andExpect(content().contentTypeCompatibleWith(mediaType));
     }
 
+    @ParameterizedTest(name = "[{index}] contentType {0}")
+    @MethodSource("getContentTypes")
+    void streamByProteomeIdAllContentType(MediaType mediaType) throws Exception {
```
LeonardoGonzales, line 559: Do you think we should extract all proteome stream tests to another IT class UniParcProteomeStreamControllerIT.java?

uniparc-rest/src/main/java/org/uniprot/api/uniparc/controller/UniParcController.java
```diff
@@ -377,6 +372,71 @@ public ResponseEntity<MessageConverterContext<UniParcEntry>> searchByProteomeId(
...
+    @Operation(
+            hidden = true,
+            summary = PROTEOME_UPID_STREAM_UNIPARC_OPERATION,
+            description = PROTEOME_UPID_STREAM_UNIPARC_OPERATION_DESC,
+            responses = {
+                @ApiResponse(
+                        content = {
+                            @Content(
+                                    mediaType = APPLICATION_JSON_VALUE,
+                                    array =
```
LeonardoGonzales, line 398: Can you please merge with master to also reflect the recent changes about OpenApi?
                            @Content(
                                    mediaType = APPLICATION_JSON_VALUE,
                                    schema = @Schema(implementation = StreamResult.class)),

PR #509, opened 2024-10-24
Support facets on uniparc cross reference

ahmadshadab:


uniparc-common/src/main/java/org/uniprot/api/uniparc/common/repository/store/crossref/UniParcCrossReferenceFacetConfig.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    List<FacetItem> getFacetItemStatuses(List<UniParcCrossReference> crossReferences) {
+        Map<String, Long> statusCount =
+                crossReferences.stream()
+                        .collect(
+                                Collectors.groupingBy(
+                                        crossRef -> crossRef.isActive() ? ACTIVE_STR : INACTIVE_STR,
+                                        Collectors.counting()));
+        Map<String, String> labelNameMap =
```
LeonardoGonzales, line 132: I do not think we need this Map, right? it can be simplified, right?

uniparc-common/src/main/java/org/uniprot/api/uniparc/common/repository/store/crossref/UniParcCrossReferenceFacetConfig.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    List<FacetItem> getFacetItemOrganisms(List<UniParcCrossReference> crossReferences) {
+        // taxonId to count map
+        Map<Long, Long> taxonIdCount =
+                crossReferences.stream()
+                        .filter(crossRef -> Objects.nonNull(crossRef.getOrganism()))
+                        .collect(
+                                Collectors.groupingBy(
+                                        crossRef -> crossRef.getOrganism().getTaxonId(),
```
LeonardoGonzales, line 147: Do you think here that we could return an Organism object and this way we would not need taxonIdName map?

LeonardoGonzales (approved):

PR #520, opened 2024-11-26
improve job creation

supun-ebi:


async-download-rest/src/main/java/org/uniprot/api/async/download/messaging/consumer/MessageConsumer.java
```diff
@@ -100,9 +98,10 @@ public void onMessage(Message message) {
...
     }
 
-    protected boolean isConsumedBefore(R downloadJob) {
-        return fileHandler.areAllFilesPresent(downloadJob.getId())
-                && ERROR != downloadJob.getStatus();
+    private static <R extends DownloadJob> boolean isJobStillRunning(R downloadJob) {
+        return downloadJob.getStatus() == RUNNING
+                || downloadJob.getStatus() == PROCESSING
+                || downloadJob.getStatus() == UNFINISHED;
     }
```
ahmadshadab, line 105: isJobStillRunning doesn't capture the intent of checking that if the job is being run by another consumer/thread. We can use something like isJobRunning, isJobCurrentlyRunning, isJobRunningByOtherConsumer.
Do we need this method to be static? I think, we can remove it. I think, we can remove generics <R extends DownloadJob> because this method is being called from this onMessage method of this class only.

async-download-rest/src/main/java/org/uniprot/api/async/download/service/JobService.java
```diff
@@ -13,7 +14,15 @@ public JobService(DownloadJobRepository<R> downloadJobRepository) {
...
+    public R create(R downloadJob) {
+        downloadJobRepository
+                .findById(downloadJob.getId())
+                .ifPresent(
+                        dj -> {
+                            throw new IllegalDownloadJobSubmissionException(
+                                    downloadJob.getId(),
+                                    "A job with id %s is already present.".formatted(dj.getId()));
+                        });
         return downloadJobRepository.save(downloadJob);
```
ahmadshadab, line 26: You can change this message as @LeonardoGonzales suggested.

ahmadshadab (commented):
Please merge after addressing those minor comments.

PR #525, opened 2024-12-11
added statistics_category column to attribute_query table

supun-ebi:


support-data-rest/src/test/resources/init.sql
```diff
@@ -44,7 +44,8 @@ ALTER TABLE uniprotkb_statistics_entry
 CREATE TABLE attribute_query
 (
     id   SERIAL PRIMARY KEY,
-    attribute_name  varchar(64),
+    attribute_name  varchar(64) NOT NULL,
+    statistics_category_id int8 NOT NULL,
     query varchar(64)
 );
```
ahmadshadab, line 50: Minor comment - To mimick the actual table, we can also add unique key constraint on attribute_name and statistics_category_id in H2

ahmadshadab (approved):
Do we not need any migration sql script for existing data? Will we copy the data along with table definition from stage to prod?

PR #526, opened 2024-12-12
Trm 32237 accessions endpoint curoff results

supun-ebi:


common-rest/src/main/java/org/uniprot/api/rest/service/StoreStreamerSearchService.java
```diff
@@ -102,6 +102,9 @@ public Stream<String> streamIdsForDownload(StreamRequest request) {
     }
 
     public QueryResult<R> getByIds(IdsSearchRequest idsRequest) {
+        if (idsRequest.getSize() == null) {
```
LeonardoGonzales, line 105: Can you please move this logic to BasicRequestConverter.createSearchIdsSolrRequest? this is the new place that we handle with logics related with IdsSearchRequest. Can you please add a Unit Test for it as well?

ahmadshadab (approved):

PR #535, opened 2025-02-06
Trm 28087 support greek letters solr search

supun-ebi:


common-rest/src/main/java/org/uniprot/api/rest/service/query/processor/UniProtSpecialCharacterQueryNodeProcessor.java
```diff
@@ -0,0 +1,157 @@
...
+    @Override
+    protected QueryNode postProcessNode(QueryNode node) throws QueryNodeException {
+        if (node.isLeaf() && node instanceof FieldQueryNode fieldQueryNode) {
+            try (UniProtSpecialCharacterFilter uniProtSpecialCharacterFilter =
+                         new UniProtSpecialCharacterFilter(new CharSequenceReader(fieldQueryNode.getText()))) {
+                uniProtSpecialCharacterFilter.read();
+                uniProtSpecialCharacterFilter.reset();
+                fieldQueryNode.setText(IOUtils.toString(uniProtSpecialCharacterFilter));
+            } catch (Exception e) {
+                throw new QueryNodeException(e);
```
ahmadshadab, line 30: We should catch specific exception IOEXception instead of broader one Exception. 
Could you please add a test for the exception case ? What happens to the search and search query in case of exception?

common-rest/src/main/java/org/uniprot/api/rest/service/query/processor/UniProtSpecialCharacterQueryNodeProcessor.java
```diff
@@ -0,0 +1,157 @@
...
+    protected QueryNode preProcessNode(QueryNode node) {
+        return node;
+    }
+
+    @Override
+    protected QueryNode postProcessNode(QueryNode node) throws QueryNodeException {
+        if (node.isLeaf() && node instanceof FieldQueryNode fieldQueryNode) {
+            try (UniProtSpecialCharacterFilter uniProtSpecialCharacterFilter =
+                         new UniProtSpecialCharacterFilter(new CharSequenceReader(fieldQueryNode.getText()))) {
+                uniProtSpecialCharacterFilter.read();
```
ahmadshadab, line 26: Could you please check if we can avoid creating a new objectof UniProtSpecialCharacterFilter for every QueryNode

common-rest/src/main/java/org/uniprot/api/rest/service/query/processor/UniProtQueryNodeProcessorPipeline.java
```diff
@@ -27,5 +27,6 @@ public UniProtQueryNodeProcessorPipeline(UniProtQueryProcessorConfig queryProces
         add(new UniProtPointRangeQueryNodeProcessor());
         add(new UniProtTermRangeQueryNodeProcessor());
         add(new UniProtDefaultFieldQueryNodeProcessor());
+        add(new UniProtSpecialCharacterQueryNodeProcessor());
     }
```
ahmadshadab, line 31: Add timiing and see how much processors are taking before and after special character node processor

common-rest/src/test/java/org/uniprot/api/rest/service/query/UniProtQueryProcessorTest.java
```diff
@@ -236,6 +236,14 @@ void handleCamelCaseFieldSearch() {
...
     }
 
+    @Test
+    void handleGreekSymbolValueSearch() {
+        String query = "field:other_value"+"\u03B1";
+        when(searchFieldConfig.getSearchFieldNames()).thenReturn(Set.of("field"));
+        String processedQuery = processor.processQuery(query);
+        assertThat(processedQuery, is("field:other_valuealpha"));
+    }
+
```
ahmadshadab, line 246: Add more positive and negative unit tests.
Add integration test at controller level as well.

common-rest/src/main/java/org/uniprot/api/rest/service/query/processor/UniProtQueryNodeProcessorPipeline.java
supun-ebi: 3 milis

common-rest/src/main/java/org/uniprot/api/rest/service/query/processor/UniProtSpecialCharacterQueryNodeProcessor.java
supun-ebi: no, its final

ahmadshadab (approved):
Just add a test in UniProtQueryProcessorTest.java with empty query if not there.

PR #538, opened 2025-02-24
Trm 32283 uniparc xref add params

supun-ebi:


uniparc-common/src/main/java/org/uniprot/api/uniparc/common/service/request/UniParcDatabasesRequest.java
```diff
@@ -27,6 +27,12 @@ public class UniParcDatabasesRequest extends UniParcGetByIdRequest implements Se
     @ValidReturnFields(uniProtDataType = UniProtDataType.UNIPARC_CROSSREF)
     private String fields;
 
+    @Parameter(description = XREF_ID_UNIPARC_DESCRIPTION, example = XREF_ID_UNIPARC_EXAMPLE)
+    private String id;
+
+    @Parameter(description = XREF_INCLUDE_SOURCES_UNIPARC_DESCRIPTION, example = XREF_INCLUDE_SOURCES_UNIPARC_EXAMPLE)
+    private boolean includeSources;
```
ahmadshadab, line 34: Hide this from doc

uniparc-common/src/main/java/org/uniprot/api/uniparc/common/service/filter/UniParcDatabaseIdFilter.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.uniprot.core.uniparc.UniParcCrossReference;
+
+import java.util.Objects;
+import java.util.function.BiFunction;
+
+/**
+ * @author sahmad
+ * @created 11/08/2020
+ */
```
ahmadshadab, line 11: Remove my name

uniparc-common/src/main/java/org/uniprot/api/uniparc/common/service/light/UniParcCrossReferenceService.java
```diff
@@ -154,6 +172,12 @@ private Stream<UniParcCrossReference> getFilteredCrossReferences(
                 .filter(xref -> filterCrossReference(xref, request));
     }
 
+    private Stream<UniParcCrossReference> getFilteredDatabaseCrossReferences(
+            UniParcEntryLight uniParcEntryLight, UniParcDatabasesRequest request) {
+        return getFilteredCrossReferences(uniParcEntryLight,request)
+                .filter(xref -> new UniParcDatabaseIdFilter().apply(xref, request.getId()));
+    }
```
ahmadshadab, line 179: No need to create filter every time. 
Create only one object of type new UniParcDatabaseIdFilter() and reuse it

uniparc-common/src/main/java/org/uniprot/api/uniparc/common/service/light/UniParcCrossReferenceService.java
```diff
@@ -101,19 +105,32 @@ public QueryResult<UniParcCrossReference> getCrossReferencesByUniParcId(
...
     }
 
+    private UniParcCrossReference hideSources(UniParcCrossReference uniParcCrossReference) {
+            if (uniParcCrossReference.getProperties().stream().noneMatch(prop -> SOURCES.equals(prop.getKey()))) {
+                return uniParcCrossReference;
+            }
+            return UniParcCrossReferenceBuilder
+                    .from(uniParcCrossReference).propertiesSet((uniParcCrossReference.getProperties().stream()
+                            .filter(prop -> !SOURCES.equals(prop.getKey())).toList())).build();
+    }
```
ahmadshadab, line 132: We can do it in one iteration in all case.
Get list of filtered properties if it is same as original properties count then we can return original otherwise set the properties

uniparc-common/src/main/java/org/uniprot/api/uniparc/common/service/light/UniParcCrossReferenceService.java
ahmadshadab: I think, we should ignore case while comparing. q12345 should match Q12345

uniparc-common/src/test/java/org/uniprot/api/uniparc/common/service/light/UniParcCrossReferenceServiceTest.java
```diff
@@ -71,6 +74,68 @@ void testGetCrossReferencesByUniParcId_Success() {
...
+                service.getCrossReferencesByUniParcId(uniParcId, request);
+
+        // then
+        assertNotNull(result);
+        List<UniParcCrossReference> content = result.getContent().toList();
+        assertTrue(content.stream().map(UniParcCrossReference::getProperties).flatMap(List::stream)
+                .map(Property::getKey).anyMatch(SOURCES::equals));
+        assertEquals(10, content.size());
+        assertNotNull(result.getPage());
+    }
```
ahmadshadab, line 137: We should assert the sources values also

uniparc-common/src/main/java/org/uniprot/api/uniparc/common/service/light/UniParcCrossReferenceService.java
supun-ebi: its basically the same what i'm doing here? any advantage of comparing based on the count over nonMatch()?

supun-ebi: got it

ahmadshadab (approved):

PR #555, opened 2025-06-27
make includeIsoform field optional

ahmadshadab:


mapto-rest/src/test/java/org/uniprot/api/mapto/controller/BaseMapToControllerIT.java
```diff
@@ -158,12 +159,24 @@ public void stopCluster() throws Exception {
...
         ResultActions response = callRun(query, includeIsoform);
 
+        return verifyRunResponseAndGetJobId(response);
+    }
+
+    private static @NotNull String verifyRunResponseAndGetJobId(ResultActions response)
+            throws Exception {
         // then
-        response.andDo(log())
+        response.andDo(print())
```
supun-ebi, line 179: remove the print()

mapto-common/src/main/java/org/uniprot/api/mapto/common/model/MapToJob.java
```diff
@@ -35,7 +36,7 @@ public class MapToJob implements Serializable {
     @JsonSerialize(using = LocalDateTimeSerializer.class)
     private LocalDateTime created;
 
-    private List<String> targetIds;
+    private List<String> targetIds = new ArrayList<>();
```
supun-ebi, line 39: why to have a init here? and with an arraylist?

supun-ebi (approved):

PR #561, opened 2025-07-29
rename organelle as encoded_in

ahmadshadab:


supun-ebi (approved):
please check the related tests

PR #562, opened 2025-08-01
Trm 27650 Provide support for uniref members stream

supun-ebi:


ahmadshadab (approved):
LGTM. Just test that it is generating the correct Swagger doc.

PR #566, opened 2025-08-11
remove unwated test verification

supun-ebi:


mapto-common/src/test/java/org/uniprot/api/mapto/common/service/MapToJobSubmissionServiceTest.java
```diff
@@ -92,7 +92,6 @@ void submit_alreadyExistingJob() {
 
         JobSubmitResponse jobSubmitResponse = mapToJobSubmissionService.submit(mapToJobRequest);
         verify(jobTaskExecutor, never()).execute(any());
-        verify(mapToJobService).updateUpdated(ID);
```
Vishal-Joshi, line 95: Hi Supun,

I don’t know anything about this codebase. So, I am not sure why the test is failing? 
I see [MapToJobSubmissionService.java's](https://github.com/ebi-uniprot/uniprot-rest-api/blob/main/mapto-common/src/main/java/org/uniprot/api/mapto/common/service/MapToJobSubmissionService.java) that submit method is not calling mapToJobService at all when mapToJobExists is true.

Perhaps to be 100% safe, you should verify for
`verifyNoInteractions(mapToJobService);`

Vishal-Joshi (changes requested):

Vishal-Joshi (approved):
had a chat with Supun on chat. This `mapToJobService#updateUpdated` is not required anymore due to recent storage technology change. Approving it therefore, in order to unblock him.

PR #604, opened 2026-03-26
increase the boost for reviewed uniprotkb entries

supun-ebi:


ahmadshadab (commented):
Please add few test cases. Unit and may be integ tests

ahmadshadab (approved):

PR #607, opened 2026-04-15
change proteome type example value for integ tests

supun-ebi:


ahmadshadab (commented):
Could you add test to search by proteome_type and verify returned type in the response and a negative test like lowercase value doesn't work.

ahmadshadab (approved):

PR #622, opened 2026-06-16
add support for different formats for precompute get by id

supun-ebi:


ahmadshadab (approved):
@supun-ebi - I’m approving this. But, the test class does not follow the pattern used by our other integration test classes We should add a test to verify the behaviour when a user requests a format that is not supported.

PR #623, opened 2026-06-17
use kb solr client for the literature repo

supun-ebi:


ahmadshadab (commented):
I was thinking our ITs should also be aware of 2 solr clusters otherwise the code will work locally but not in k8s.
We should file a ticket and work on it.

ahmadshadab:
> I was thinking our ITs should also be aware of 2 solr clusters otherwise the code will work locally but not in k8s. We should file a ticket and work on it.

Feel free to file a separate ticket for the above tasks and you can merge this if you have tested it in dev

ahmadshadab (approved):

PR #624, opened 2026-06-30
feat: TRM-34151: Add JSON based stream endpoint for precomputed annotations by proteome id

rajind-ebi:
This PR is to to add JSON based stream endpoint for precomputed annotations by proteome id.

- Note that this PR depends on the schema change at https://github.com/ebi-uniprot/uniprot-store/pull/448
- PR for the config property changes were added at https://gitlab.ebi.ac.uk/uniprot/deployment/unp.ci.api.k8s/-/merge_requests/131

### DIFF
--- a/uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationRequestConverterImpl.java
+++ b/uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationRequestConverterImpl.java
@@ -9,11 +9,8 @@ import org.uniprot.api.rest.service.query.processor.UniProtQueryProcessorConfig;
 import org.uniprot.api.rest.service.request.BasicRequestConverter;
 import org.uniprot.api.rest.service.request.RequestConverterConfigProperties;
 import org.uniprot.api.rest.service.request.RequestConverterImpl;
-import org.uniprot.store.search.SolrQueryUtil;
 
 public class PrecomputedAnnotationRequestConverterImpl extends RequestConverterImpl {
-    private static final String TAXONOMY_ID = "taxonomy_id";
-
     private final BasicRequestConverter basicConverter;
 
     public PrecomputedAnnotationRequestConverterImpl(
@@ -44,7 +41,6 @@ public class PrecomputedAnnotationRequestConverterImpl extends RequestConverterI
         SolrRequest.SolrRequestBuilder builder =
                 basicConverter.createSearchSolrRequest(
                         precomputedAnnotationSearchByProteomeRequest);
-        builder.query(createQuery(precomputedAnnotationSearchByProteomeRequest));
         return builder.build();
     }
 
@@ -56,11 +52,5 @@ public class PrecomputedAnnotationRequestConverterImpl extends RequestConverterI
                 basicConverter.createStreamSolrRequest(
                         precomputedAnnotationStreamByProteomeRequest);
         return builder.build();
-    }
-
-    private String createQuery(PrecomputedAnnotationSearchByProteomeRequest request) {
-        String escapedTaxonomyId =
-                SolrQueryUtil.escapeSpecialCharacters(request.getTaxonomyId().strip());
-        return TAXONOMY_ID + ":" + escapedTaxonomyId;
     }
 }
--- a/uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationRequestConverterImpl.java
+++ b/uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationRequestConverterImpl.java
@@ -3,6 +3,7 @@ package org.uniprot.api.uniprotkb.common.service.precomputed;
 import org.uniprot.api.common.repository.search.SolrQueryConfig;
 import org.uniprot.api.common.repository.search.SolrRequest;
 import org.uniprot.api.rest.request.SearchRequest;
+import org.uniprot.api.rest.request.StreamRequest;
 import org.uniprot.api.rest.search.AbstractSolrSortClause;
 import org.uniprot.api.rest.service.query.processor.UniProtQueryProcessorConfig;
 import org.uniprot.api.rest.service.request.BasicRequestConverter;
@@ -47,7 +48,24 @@ public class PrecomputedAnnotationRequestConverterImpl extends RequestConverterI
         return builder.build();
     }
 
+    @Override
+    public SolrRequest createStreamSolrRequest(StreamRequest request) {
+        PrecomputedAnnotationStreamByProteomeRequest precomputedAnnotationStreamByProteomeRequest =
+                (PrecomputedAnnotationStreamByProteomeRequest) request;
+        SolrRequest.SolrRequestBuilder builder =
+                basicConverter.createStreamSolrRequest(
+                        precomputedAnnotationStreamByProteomeRequest);
+        builder.query(createQuery(precomputedAnnotationStreamByProteomeRequest));
+        return builder.build();
+    }
+
     private String createQuery(PrecomputedAnnotationSearchByProteomeRequest request) {
+        String escapedTaxonomyId =
+                SolrQueryUtil.escapeSpecialCharacters(request.getTaxonomyId().strip());
+        return TAXONOMY_ID + ":" + escapedTaxonomyId;
+    }
+
+    private String createQuery(PrecomputedAnnotationStreamByProteomeRequest request) {
         String escapedTaxonomyId =
                 SolrQueryUtil.escapeSpecialCharacters(request.getTaxonomyId().strip());
         return TAXONOMY_ID + ":" + escapedTaxonomyId;
--- a/uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationRequestConverterImpl.java
+++ b/uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationRequestConverterImpl.java
@@ -55,17 +55,10 @@ public class PrecomputedAnnotationRequestConverterImpl extends RequestConverterI
         SolrRequest.SolrRequestBuilder builder =
                 basicConverter.createStreamSolrRequest(
                         precomputedAnnotationStreamByProteomeRequest);
-        builder.query(createQuery(precomputedAnnotationStreamByProteomeRequest));
         return builder.build();
     }
 
     private String createQuery(PrecomputedAnnotationSearchByProteomeRequest request) {
-        String escapedTaxonomyId =
-                SolrQueryUtil.escapeSpecialCharacters(request.getTaxonomyId().strip());
-        return TAXONOMY_ID + ":" + escapedTaxonomyId;
-    }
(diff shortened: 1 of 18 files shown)


uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationStreamByProteomeRequest.java
```diff
@@ -0,0 +1,66 @@
...
+    @ValidReturnFields(uniProtDataType = UniProtDataType.PRECOMPUTED_ANNOTATION)
+    private String fields;
+
+    @Parameter(hidden = true)
+    private String format;
+
+    @Override
+    public String getQuery() {
+        return TAXONOMY_ID_STR + ":" + this.upId;
+    }
```
ahmadshadab, line 63: Based on our discussion, we agree that this query is wrong. We are using wrong field name for the upId. Remove it if we don't need it.

uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationRequestConverterImpl.java
```diff
@@ -47,9 +48,26 @@ public SolrRequest createSearchSolrRequest(SearchRequest request) {
...
                 SolrQueryUtil.escapeSpecialCharacters(request.getTaxonomyId().strip());
         return TAXONOMY_ID + ":" + escapedTaxonomyId;
     }
+
+    private String createQuery(PrecomputedAnnotationStreamByProteomeRequest request) {
+        String escapedTaxonomyId =
+                SolrQueryUtil.escapeSpecialCharacters(request.getTaxonomyId().strip());
+        return TAXONOMY_ID + ":" + escapedTaxonomyId;
+    }
 }
```
ahmadshadab, line 56: createSearchSolrRequest/createStreamSolrRequest and the two createQuery overloads are duplicated logic. Extract a single createQuery(String taxonomyId) helper and call it from both

uniprotkb-common/src/test/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationRequestConverterImplTest.java
```diff
@@ -30,6 +30,23 @@ void createsSolrQueryFromResolvedTaxonomyId() {
...
+        SolrRequest solrRequest = converter.createStreamSolrRequest(request);
+
+        assertEquals("taxonomy_id:9606", solrRequest.getQuery());
+        assertEquals("accession uniparc taxonomy_id", solrRequest.getQueryField());
+        assertEquals(100, solrRequest.getRows());
+        assertEquals(Integer.MAX_VALUE, solrRequest.getTotalRows());
+        assertEquals("accession", solrRequest.getSorts().get(0).getItem());
+        assertEquals(SolrQuery.ORDER.desc, solrRequest.getSorts().get(0).getOrder());
+    }
+
```
ahmadshadab, line 50: What is the default sort order? We generally return result sorted by accession. Check other API and add default sorting behaviour with a test here.

uniprotkb-common/src/test/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedUniProtKBEntryServiceTest.java
```diff
@@ -127,9 +133,69 @@ void searchReturnsOnlyEntriesAvailableFromStoreClient() {
...
     private PrecomputedUniProtKBEntryService service() {
         return new PrecomputedUniProtKBEntryService(
-                storeClient, repository, proteomeTaxonomyResolver, requestConverter);
+                storeClient,
+                repository,
+                proteomeTaxonomyResolver,
+                requestConverter,
+                storeStreamer,
+                solrIdStreamer);
     }
```
ahmadshadab, line 202: Add negative test what if proteome id doesn't exist.

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/PrecomputedUniProtKBController.java
```diff
@@ -143,6 +143,34 @@ public ResponseEntity<MessageConverterContext<UniProtKBEntry>> searchByProteomeI
...
+        MediaType contentType = getAcceptHeader(request);
+        setBasicRequestFormat(streamRequest, request);
+        return super.stream(
+                () -> precomputedUniProtKBEntryService.streamByProteomeId(streamRequest),
+                streamRequest,
+                contentType,
+                request);
+    }
+
     @Override
```
ahmadshadab, line 178: Test accessing the endpoint which generates openapi doc is still generating openapi doc json file properly. Sometimes, adding a new endpoint causes regression.

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/PrecomputedUniProtKBStreamControllerIT.java
```diff
@@ -0,0 +1,234 @@
...
+
+        UniProtKBEntry uniProtKBEntry = uniProtKBEntryBuilder.build();
+        PrecomputedAnnotationDocument precomputedAnnotationDocument =
+                precomputedAnnotationEntryToDocumentMapper.call(uniProtKBEntry);
+
+        cloudSolrClient.addBean(
+                SolrCollection.precomputedannotation.name(), precomputedAnnotationDocument);
+        precomputedAnnotationStoreClient.saveEntry(uniProtKBEntry);
+    }
+}
```
ahmadshadab, line 349: Can you add few negative test cases like -
1. Unsupported format requested.
2. Stream by non existing proteome id

Check uniprotkb stream api if we support default sorting. If yes, add assertion to verify sorting order.

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/UniProtKBStreamControllerIT.java
```diff
@@ -113,6 +113,7 @@ class UniProtKBStreamControllerIT extends AbstractStreamControllerIT {
     private SolrClient solrClient;
 
     @Autowired private FacetTupleStreamTemplate facetTupleStreamTemplate;
+    @Qualifier("uniProtKBTupleStream")
     @Autowired private TupleStreamTemplate tupleStreamTemplate;
```
ahmadshadab: This qualifier could cause issue in other test. Please the entire test suits of ITs and UTs in all rest modules.

uniprotkb-rest/src/main/java/org/uniprot/api/uniprotkb/controller/PrecomputedUniProtKBController.java
rajind-ebi: Thanks for the heads up. I've tested the docs URL (http://localhost:8095/uniprot/api/uniprotkb/api/docs) locally and it looks good. I have attached the output for reference here.

[docs.json](https://github.com/user-attachments/files/29700214/docs.json)

uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationStreamByProteomeRequest.java
rajind-ebi: Yes. Agreed. I based this on the existing search request `PrecomputedAnnotationSearchByProteomeRequest` that apparently have the same issue.

I've updated both places to be like follows. This avoids us having to override the query after `basicConverter.createStreamSolrRequest` call and hence we do not need to maintain a custom `createQuery` function as well.

```
    @Override
    public String getQuery() {
        validState(Utils.notNullNotEmpty(taxonomyId), "Taxonomy Id should not be null or empty");
        return TAXONOMY_ID_STR + ":" + taxonomyId;
    }
```

Hope this is good.

uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationRequestConverterImpl.java
rajind-ebi: As explained in this comment https://github.com/ebi-uniprot/uniprot-rest-api/pull/624#discussion_r3529276011, `createQuery` is no longer needed, so I have removed it.

uniprotkb-common/src/test/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedAnnotationRequestConverterImplTest.java
rajind-ebi: Thanks. Default sorting is `accession asc`. As discussed I have added another test case for `accession desc`.

uniprotkb-common/src/test/java/org/uniprot/api/uniprotkb/common/service/precomputed/PrecomputedUniProtKBEntryServiceTest.java
rajind-ebi: Thanks. Added a negative case which returns 404.

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/PrecomputedUniProtKBStreamControllerIT.java
rajind-ebi: Thanks. Added these test cases.

uniprotkb-rest/src/test/java/org/uniprot/api/uniprotkb/controller/UniProtKBStreamControllerIT.java
rajind-ebi: Good callout. I've checked other places add added the qualifier as necessary. I've also run full IT tests  with `mvn clean verify -Pintegration-tests` (and also unit tests `mvn clean test`) and seems to be good now.

uniprotkb-rest/src/main/resources/application.properties
```diff
@@ -140,4 +150,8 @@ server.tomcat.max-connections=1000
 ################################## Rest endpoint ################################
 unisave.rest.endpoint=rest.uniprot.org/unisave/
 
-max.solr.or.clause.count=50
\ No newline at end of file
+max.solr.or.clause.count=50
+
+#logging.level.root=DEBUG
+logging.level.org.springdoc=TRACE
+logging.level.org.springframework.web.servlet.mvc.method.annotation=TRACE
```
ahmadshadab: Remove these logging, it might fill the log file.

ahmadshadab (changes requested):

uniprotkb-rest/src/main/resources/application.properties
rajind-ebi: Oh. I didn't realize that these were committed. I was using this to check a few things on integration tests. Removed them. Thanks.

ahmadshadab (approved):

PR #628, opened 2026-07-15
fix: TRM-34245: Fix redisson client config loading issue due to jackson and snakeyaml mismatch

rajind-ebi:
`redis.yaml` is read during application startup to initiate redisson client and this was failing due to version mismatch of  jackson and snakeyaml libraries. As of now we do not have a test that uses the full application context hence, these issues are not caught. Raising this PR first to see if we encounter other issues on dev.

A separate PR will be raised to try and use full application context as a test.


ahmadshadab (commented):
Did you test by running the application locally? Are they starting cleanly?

ahmadshadab (approved):

PR #630, opened 2026-07-17
Solr9 shadow log stream

ahmadshadab:
Please see TRM-34254 for sample logs.
We need to do hotfix to capture the log for search and stream. Search logging is already deployed.
But now i have done some improvement. Log only 10% of the requests.

### DIFF
--- a/uniprotkb-common/src/test/java/org/uniprot/api/uniprotkb/common/repository/search/UniprotQueryRepositoryTest.java
+++ b/uniprotkb-common/src/test/java/org/uniprot/api/uniprotkb/common/repository/search/UniprotQueryRepositoryTest.java
@@ -1,0 +1,213 @@
+package org.uniprot.api.uniprotkb.common.repository.search;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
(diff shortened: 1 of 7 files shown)


uniprotkb-common/src/test/java/org/uniprot/api/uniprotkb/common/repository/search/UniprotQueryRepositoryTest.java
```diff
@@ -0,0 +1,213 @@
...
+        assertTrue(latch.await(5, TimeUnit.SECONDS));
+        verify(tupleStream).open();
+        verify(tupleStream).close();
+    }
+
+    @Test
+    void testShadowSearchPageFailure() throws Exception {
+        SolrRequest request = SolrRequest.builder().query("query").build();
+        JsonQueryRequest jsonQueryRequest = mock(JsonQueryRequest.class);
+        lenient().when(requestConverter.toJsonQueryRequest(request)).thenReturn(jsonQueryRequest);
```
supun-ebi, line 155: why do you want to use lenient inside a method?

supun-ebi (approved):

PR #632, opened 2026-07-22
feat: TRM-34196 Setup test container based Solr for integration test and apply the changes to uniprotkb-rest

rajind-ebi:
This PR sets up test container based Solr for integration tests and applies the changes to uniprotkb-rest. Note that since other modules integration tests are not updated yet. This will break other integration tests. I am raising this changes for early review and I am working on applying the same to other modules.

I will raise them as a separate PR to keep PRs smaller.

Refer readme for more details - https://github.com/ebi-uniprot/uniprot-rest-api/blob/TRM-34196-setup-test-container-based-solr-for-it-tests/README.md#enabledisable-the-solr-testcontainer

Make changes are on below files. 
- [HostAwareSolrContainer](https://github.com/ebi-uniprot/uniprot-rest-api/blob/TRM-34196-setup-test-container-based-solr-for-it-tests/common-rest/src/main/java/org/uniprot/api/common/test/HostAwareSolrContainer.java)
- [SolrTestContainerSetup](https://github.com/ebi-uniprot/uniprot-rest-api/blob/TRM-34196-setup-test-container-based-solr-for-it-tests/common-rest/src/main/java/org/uniprot/api/common/test/SolrTestContainerSetup.java)
- [AbstractStreamControllerIT](https://github.com/ebi-uniprot/uniprot-rest-api/blob/TRM-34196-setup-test-container-based-solr-for-it-tests/common-rest/src/test/java/org/uniprot/api/rest/controller/AbstractStreamControllerIT.java)


common-rest/src/main/java/org/uniprot/api/rest/respository/RepositoryConfig.java
```diff
@@ -37,7 +37,6 @@ public UniProtKBRepositoryConfigProperties uniProtKBRepositoryConfigProperties()
     }
 
     @Bean
-    @Profile("live")
     public HttpClient httpClient(RepositoryConfigProperties config) {
```
supun-ebi, line 40: Did you check the consequences of removing the profile here?

uniprotkb-common/src/main/java/org/uniprot/api/uniprotkb/common/repository/store/ResultsConfig.java
```diff
@@ -45,7 +45,6 @@
 public class ResultsConfig {
 
     @Bean("uniProtKBSolrClient")
```
supun-ebi, line 47: I guess this is the solr client in prod (live) code, please check whether it works after this change. (not in tests)

supun-ebi:
Overall, it looks ok for me; however, please check the consequences of removing the live solrClient (in the live environment).
Also, please verify the output of all the IT classes that are affected.

common-rest/src/test/java/org/uniprot/api/common/test/HostAwareSolrContainer.java
```diff
@@ -0,0 +1,68 @@
...
+
+        for (int i = 0; i < commandParts.length; i++) {
+            if (LOCALHOST.equals(commandParts[i])) {
+                commandParts[i] = advertisedHost;
+            }
+        }
+
+        setCommandParts(commandParts);
+    }
+}
```
ahmadshadab, line 68: I am not sure why do we need test related class in src/main/java. Same comment for SolrTestContainerSetup.

common-rest/src/test/resources/META-INF/spring.factories
```diff
@@ -0,0 +1 @@
+org.springframework.boot.env.EnvironmentPostProcessor=org.uniprot.api.common.test.SolrTestContainerSetup
```
ahmadshadab, line 1: It isn't scoped to test. Will it not load for prod also?

common-rest/src/main/resources/security.json
```diff
@@ -0,0 +1,21 @@
+{
```
ahmadshadab, line 1: Same as SolrTestContainerSetup/HostAwareSolrContainer, i am not comfortable with adding test related stuff in prod jar.

common-rest/src/test/java/org/uniprot/api/rest/controller/AbstractStreamControllerIT.java
```diff
@@ -66,42 +66,64 @@ public abstract class AbstractStreamControllerIT {
...
+        }
+    }
+
+    private void deleteCollectionIfExists(String collection) {
+        try {
+            CollectionAdminRequest.deleteCollection(collection).process(getSolrClient());
+            log.info("Deleted existing collection '{}'", collection);
+        } catch (Exception e) {
+            // Solr throws if the collection doesn't exist yet
+            log.debug("Collection '{}' did not exist, nothing to delete", collection);
```
ahmadshadab, line 126: I would suggest check if this exception is because of missing collection otherwise we should let it fail.

common-rest/src/test/java/org/uniprot/api/rest/MockSolrClientConfig.java
```diff
@@ -0,0 +1,41 @@
...
+    @Primary
+    public SolrClient solrClient() throws URISyntaxException {
+        return mock(SolrClient.class);
+    }
+
+    @Bean
+    @Primary
+    public HttpClient httpClient() {
+        return mock(HttpClient.class);
+    }
```
ahmadshadab, line 42: We generally use org.apache.http.client.HttpClient but here you are using java.net.http.HttpClient. Is it intentional?

common-rest/pom.xml
```diff
@@ -203,6 +203,11 @@
 			<artifactId>testcontainers-postgresql</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.testcontainers</groupId>
+			<artifactId>testcontainers-solr</artifactId>
+			<scope>provided</scope>
+		</dependency>
```
ahmadshadab, line 211: I would suggest try to start/deploy springboot application and see if it is starting cleanly.

ahmadshadab (commented):
General comment

- Why do we need to

common-rest/src/test/java/org/uniprot/api/common/test/HostAwareSolrContainer.java
rajind-ebi: Good point. I did not realize earlier that we have `test-jar` built for `common-rest`. Since we have `test-jar` built for `common-rest`, we can move these to test package. I've move them now.

We just have to make sure we add dependency in addition to normal `common-rest` dependency in usages.

```
		<dependency>
			<artifactId>common-rest</artifactId>
			<version>${project.parent.version}</version>
			<type>test-jar</type>
			<scope>test</scope>
		</dependency>
```

common-rest/src/test/resources/META-INF/spring.factories
rajind-ebi: I did have a safety check to this to not run in prod but on the reason to why, I did not realize earlier that we have `test-jar` built for `common-rest`. Since we have `test-jar` built for `common-rest`, we can move these to test package. I've move them now.

We just have to make sure we add dependency in addition to normal `common-rest` dependency in usages.

```
		<dependency>
			<artifactId>common-rest</artifactId>
			<version>${project.parent.version}</version>
			<type>test-jar</type>
			<scope>test</scope>
		</dependency>
```

common-rest/src/main/resources/security.json
rajind-ebi: Addressed. https://github.com/ebi-uniprot/uniprot-rest-api/pull/632#discussion_r3688711232

common-rest/src/test/java/org/uniprot/api/rest/controller/AbstractStreamControllerIT.java
rajind-ebi: This is not part of the test. This is more of a cleanup before we re-create collection from scratch for the subsequent tests.

In cases where the collection set if different from the prior test, we will always get "collection doesn't exist error". That is why I decided to ignore that exception. I guess these are things we have to handle since we are reusing the same solr container.

common-rest/src/test/java/org/uniprot/api/rest/MockSolrClientConfig.java
rajind-ebi: Good catch. I've addressed this.

common-rest/src/main/java/org/uniprot/api/rest/respository/RepositoryConfig.java
rajind-ebi: When profile tag is removed, the bean will be created with any profile. Since I am no longer mocking this solr client related beans, for the integration tests to run (with profile "offline"), I had to remove this. The integration tests now use the normal production bean for solr client.

I've run uniprotkb-rest locally once and it was good.

